Issue #296, opened 2014-11-14
utils.scope_to_list TypeError if scope is None


JamSquad:
Some providers (Specifically Mailchimp) don't supply a token_type or a scope with their OAuth authentication flow.

This causes a TypeError ('NoneType' object is not iterable) when trying to init the OAuth2Token class.  (Because it is wrapped in a set()).

Possible solution is to try, except TypeError around this call in tokens.py line 30

Or possibly I'm missing something entirely.

flamusdiu:
However, this doesn't fix #306 for me. If a Oauth2 server doesn't return the a scope on the response, then it assumed it changed. The Oauth2 may not return a scope in the response if it didn't change which scope was requested. More info on #306 as well. 

I am not sure why but another person I know using Google + flask-oauthlib doesn't get this error but I can create it 100% of the time.

mrwalker:
I worked around this issue for MailChimp in particular using a compliance fix at the OAuth client level in requests-oauthlib.  If you're stuck and don't have the luxury of waiting on a release, you can host the compliance fixes in your own code.

PR #297, opened 2014-11-14
Put try, except around OAuth2Token __init__

JamSquad:
Put try, except around OAuth2Token.__init__ so NoneType scopes don't raise errors.

Fix for issue #296


josephtate:
This pull request doesn't actually fix the problems people are having with google oauth2. #323 does.

PR #300, opened 2014-11-26
Revocation endpoint should return empty string, not Python 'None'

jbkkd:
References https://github.com/evonove/django-oauth-toolkit/issues/158

### DIFF
--- a/oauthlib/oauth2/rfc6749/endpoints/revocation.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/revocation.py
@@ -74,7 +74,7 @@ class RevocationEndpoint(BaseEndpoint):
         self.request_validator.revoke_token(request.token,
                                             request.token_type_hint, request)
 
-        response_body = None
+        response_body = ''
         if self.enable_jsonp and request.callback:
             response_body = request.callback + '();'
         return {}, response_body, 200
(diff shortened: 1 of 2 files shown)


thedrow:
@lepture I'll let you decide on this.

lepture:
If the response body is `''`, it means the response has no content. In this case, the HTTP code should be 204.

However, the spec defined the response with:

> The authorization server responds with HTTP status code 200 if the
   token has been revoked successfully or if the client submitted an
   invalid token.

The response code should be 200. Is it ok that a 200 response with no content? Would it be better to let the developer decide themselves?

thedrow:
I think that returning 200 OK without content is acceptable and does happen in the wild.
No idea about the spec though. As far as I recall HTTP allows 200 OK with no content.

lepture:
We do have servers that responses with wrong content type / encoding etc in the wild. "in the wild" doesn't mean it is right.

Here is a quote from the spec http://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7231.html#status.200

> Aside from responses to CONNECT, a 200 response always has a payload, though an origin server may generate a payload body of zero length. If no payload is desired, an origin server ought to send 204 (No Content) instead.

It has an uncertain statement:

>  though an origin server **may** generate a payload body of zero length

thedrow:
@lepture I'm waiting for a decision on this PR to merge.

lepture:
It failed

PR #305, opened 2015-01-13
Allow ! character in query string

artooro:
My app broke because there was a ! character in the query string. This small change fixed it for me.


thedrow:
Can you please rebase?

PR #307, opened 2015-01-18
Patch 1

flamusdiu:
Pull request for idan#306


flamusdiu:
Going to close this in favor of #297 solution (which seem more robust anyway). =D

josephtate:
This change does not contain tests. I'd vote for closing it in favor of #323, which is technically similar, but much more complete.

PR #314, opened 2015-02-10
Check for rsa_key when it's actually needed

singingwolfboy:
We might want to instantiate a RSA OAuth1 client, but not load the RSA key until later. We should move the check from where the client is initialized, to where the RSA key is actually used.


josephtate:
Can you freshen this? And maybe provide a unit test?

singingwolfboy:
@josephtate fixed!

josephtate:
LGTM pending travis run.

singingwolfboy:
@josephtate ping

josephtate:
@singingwolfboy I'm not able to merge pull requests, just pitching in to review. :( @ib-lundgren this LGTM.

PR #321, opened 2015-02-26
update docs for save_authorization_code

djmitche:
The `rtype` here was, I think, copy/pasted from above.

The `code` parameter is a dictionary containing only one key.  I suppose this has been left for future expansion, but `validate_code` only passes the code itself, so there's no need to persist more than this.

Related question: what does it mean to "associate" the code with these things?  When would I validate that association?


thedrow:
Thanks for the contribution. This will be merged in #360.

Issue #226, opened 2013-12-09
Why PyCrypto for RSA?


TkTech:
About two years ago it appears pycrypto was swapped out for python-rsa. Then at some point about a year ago pycrypto was re-introduced (this time as a soft dependency).

My questions:
- Why was this decision made?
- If python-rsa was no longer viable, why not the alternatives like tlslite?
- Why is there no easy extension mechanism for swapping out implementations?

I ask because [jira-python][] already includes [tlslite][] which is perfectly capable of doing the signing, but is pure-python and can optionally use pycrypto for acceleration. Howerver, there is no easy (or maybe there is but undocumented?) way of swapping it out without forking the code. I had to resort to monkey patching to do so:

```python
def _pure_python_rsa_sha1(base_string, rsa_private_key):
    """
    An alternative, pure-python RSA-SHA1 signing method, using
    the tlslite library.
    """
    import base64
    from tlslite.utils import keyfactory

    private_key = keyfactory.parsePrivateKey(rsa_private_key)
    signature = private_key.hashAndSign(base_string.encode('ascii'))

    # This *must* be unicode, since oauthlib refuses to work with
    # anything else.
    return unicode(base64.b64encode(signature))

# Monkey patch the built-in RSA-SHA1 signing method, since the
# package offers no proper extension mechanism.
from oauthlib.oauth1.rfc5849 import signature
signature.sign_rsa_sha1 = _pure_python_rsa_sha1
```

[jira-python]: http://jira-python.readthedocs.org/en/latest/
[tlslite]: https://pypi.python.org/pypi/tlslite

Zopieux:
@ib-lundgren Just so you know, [pyca/cryptography#1331](https://github.com/pyca/cryptography/pull/1331) has been merged.

thedrow:
Woot!

tilgovi:
Can be closed now that #322 merged.

thedrow:
@ib-lundgren ping. You can close this issue now.

PR #322, opened 2015-02-27
This patch fixes PyJWT support when pyjwt >= 0.4.0

josephtate:
Newer versions of pyjwt subsume a bunch of the PEM decoding stuff.

This fixes tests with all latest deps.

I also reformatted the PEM key strings so that they could be copied/pasted easier. PEP-8 doesn't forbid multi-line strings.

Note that the Private Keys were retrieved by running RSA.importKey(private_pem).publickey.exportKey().

### DIFF
--- a/oauthlib/common.py
+++ b/oauthlib/common.py
@@ -229,10 +229,7 @@ def generate_token(length=30, chars=UNICODE_ASCII_CHARACTER_SET):
 
 
 def generate_signed_token(private_pem, request):
-    import Crypto.PublicKey.RSA as RSA
     import jwt
-
-    private_key = RSA.importKey(private_pem)
 
     now = datetime.datetime.utcnow()
 
@@ -243,23 +240,16 @@ def generate_signed_token(private_pem, request):
 
     claims.update(request.claims)
 
-    token = jwt.encode(claims, private_key, 'RS256')
+    token = jwt.encode(claims, private_pem, 'RS256')
     token = to_unicode(token, "UTF-8")
 
     return token
 
 
-def verify_signed_token(private_pem, token):
-    import Crypto.PublicKey.RSA as RSA
+def verify_signed_token(public_pem, token):
     import jwt
 
-    public_key = RSA.importKey(private_pem).publickey()
-
-    try:
-        # return jwt.verify_jwt(token.encode(), public_key)
-        return jwt.decode(token, public_key)
-    except:
-        raise Exception
+    return jwt.decode(token, public_pem, algorithms=['RS256'])
 
 
 def generate_client_id(length=30, chars=CLIENT_ID_CHARACTER_SET):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,5 +1,11 @@
 Changelog
 =========
+Unreleased
+--------------------------------------------------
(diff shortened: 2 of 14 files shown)


josephtate:
The CI build failed because tox.ini isn't being respected. PyJWT now requires "cryptography" vs. PyCrypto.

lepture:
ping @ib-lundgren

tilgovi:
LGTM

wiliamsouza:
:+1:  LGTM

jpadilla:
Yesterday we released PyJWT [v1.0.0](https://github.com/jpadilla/pyjwt/releases/tag/1.0.0) which fixes some reported security vulnerabilities. Although this project doesn't seem to be affected, it would still be good idea to require >= 1.0.0 instead.

josephtate:
@jpadilla Thanks for the heads up

ib-lundgren:
Sorry for not having looked at this earlier. I just merged #310 which changed the travis/tox setup a bit, can you please sync and set requirements for pyjwt as needed? 

I see cryptography being installed but never imported / explicitly used. I'm guessing pyjwt does not install this implicitly and we have do it?

tilgovi:
Only if we need to test the asymmetric ciphers.

jpadilla:
PyJWT now allows you to either use pycrypto or cryptography(recommended) for asymmetric ciphers, which this project does use (RS256) to be exact. None of them are installed by default.

josephtate:
I'm working on re-merging. I'm also realizing that pycrypto could possibly be removed, and that there are docs that need to be fixed up.

ib-lundgren:
OAuth1 uses pycrypto as well. I begun a CL to change to cryptography / use either but cryptography was not quite ready at the time. Since then I've not been able to finish it up. Feel free to resume it #226 :)

josephtate:
@jpadilla @ib-lundgren Can I get a review of this? I don't know that I like the way that I'm handling oauth1 signatures sans pycrypto, but cryptography doesn't seem to have its low level signing routines polished very well.

I'll work on updating docs.

oauthlib/common.py
```diff
@@ -243,23 +240,16 @@ def generate_signed_token(private_pem, request):
...
     import jwt
 
-    public_key = RSA.importKey(private_pem).publickey()
-
-    try:
-        # return jwt.verify_jwt(token.encode(), public_key)
-        return jwt.decode(token, public_key)
-    except:
-        raise Exception
+    return jwt.decode(token, public_pem)
```
jpadilla: I'd suggest specifying the allowed algorithms for decoding, in this case it'd be something like:

```python
return jwt.decode(token, public_pem, algorithms=['RS256'])
```

josephtate:
I think I'm all done with this @ib-lundgren.

lepture:
@josephtate maybe you should rebase your commits, make it clear.

josephtate:
Welp, That was fun. I rebased, but that broke, so I re-patched master and force pushed the branch.

Summary:
This current version happens to replace #226, and accidentally includes #336 (which is minor).

ib-lundgren:
Amazing :) Well done and thanks for the help reviewing this all!

PR #323, opened 2015-02-27
Fix #306: handle token responses where no scope is provided because it has not changed

### LINKED ISSUE #306: Google OAuth2 returns no scope on authentication breaks scope test
Google doesn't seem to return an empty scope when authentication over Oauth2:

```
params = {
  "access_token" : "...",
  "token_type" : "Bearer",
  "expires_in" : 3600,
  "id_token" : "..."
}
```
This is what I get when the authentication works. It seems that https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/tokens.py#L30 contains `set(utils.scope_to_list(params.get('scope', '')))` which sets the the scope to empty because there is no scope returned from the servers. 

This fails the check for [params.scope_changed](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/parameters.py#L396) even though the scope works for Google.

To get this to work I changed `set(utils.scope_to_list(params.get('scope', '')))` to `set(utils.scope_to_list(params.get('scope', old_scope)))`.

Oh, and if you look here: http://tools.ietf.org/html/rfc6749#section-3.3. From my understanding: "If the issued access token scope is different from the one requested by the client, the authorization server MUST include the "scope" response parameter to inform the client of the actual scope granted." Which means, since the scope didn't change, then they are not required to send it back, correct?

More information about this located on my StackOverflow question: [Flask-Dance Error: Scope has changed](http://stackoverflow.com/questions/28011269/flask-dance-error-scope-has-changed)

josephtate:
Adds a test (failing), then fixes the test by handling the case where "scope" is not in the params.

### DIFF
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -27,9 +27,15 @@ class OAuth2Token(dict):
 
     def __init__(self, params, old_scope=None):
         super(OAuth2Token, self).__init__(params)
-        self._new_scope = set(utils.scope_to_list(params.get('scope', '')))
+        self._new_scope = None
+        if 'scope' in params:
+            self._new_scope = set(utils.scope_to_list(params['scope']))
         if old_scope is not None:
             self._old_scope = set(utils.scope_to_list(old_scope))
+            if self._new_scope is None:
+                # the rfc says that if the scope hasn't changed, it's optional
+                # in params so set the new scope to the old scope
+                self._new_scope = self._old_scope
         else:
             self._old_scope = self._new_scope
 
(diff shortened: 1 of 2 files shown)


josephtate:
This pull request fixes #296, #306, and (IMO) supercedes #297 and #307. It also merges cleanly with my #322.

ib-lundgren:
Looks good to me :) Cheers!

jonnor:
Is there a release with this fix in? Or a workaround available?

josephtate:
The fix is in the git master branch. You can check it out and use it
directly with pip. Unfortunately it has not been released yet.

PR #324, opened 2015-02-27
Some simple fixes for OAuth 2 base client

JonnyFunFun:
While developing a solution using the OAuth 2 client, I noticed there were some variables in base.py that were wrong.  This was causing things like preparing a token refresh request to fail due to an AttributeError.  This simply fixes two of those situations that I've encountered.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ python:
 
 install:
   - if [[ $TRAVIS_PYTHON_VERSION == '2.6' ]]; then pip install --use-mirrors unittest2; fi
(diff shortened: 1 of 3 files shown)


JonnyFunFun:
I see that this change causes tests to fail, but without these changes both prepare_authorization_request and prepare_refresh_token_request fail when using WebApplicationClient.  I will continue to mess around with this and see what it causing it, but it appears that it may be my issue internally.

thedrow:
Already in master. Thanks for your contribution.

PR #326, opened 2015-03-01
fix method params and clarify oauth_protected

CoolGoose:
I've updated a bit the docs in order to clarify which endpoint you need to use and fix the method params.
If it's ok i'll look over the docs a bit more and update it with more vendor agnostic code where possible (it's flask-ish here and there :) )


thedrow:
Thanks!

PR #328, opened 2015-03-03
Error description in InvalidRequestError should be oauth1 not oauth2

bootandy:


lepture:
failed due to https://github.com/idan/oauthlib/pull/322

thedrow:
Can you please rebase? I don't think it should fail now.

PR #335, opened 2015-03-26
Precise oauth1 client workflow in documentation

st4lk:
If i understand correctly the OAuth 1.0a, https://oauthlib.readthedocs.org/en/latest/oauth1/client.html have some inaccuracies. For example, to obtain access token and access secret we need to provide both request token and request secret, not just request secret.


ib-lundgren:
Looks good to me :) Thanks for fixing this!

PR #339, opened 2015-05-06
Add code to determine if client authentication is required for OAuth2…

weipin:
An attempt to improve OAuth2 endpoint "revocation.py", and to fix django-oauth-toolkit issues https://github.com/evonove/django-oauth-toolkit/issues/244 and https://github.com/evonove/django-oauth-toolkit/issues/242.

The "client authentication" is always required in the current method `validate_revocation_request`. There are cases that the authentication can be ignored though, like when grand type is "Resource Owner Password Credentials" and the "Client type" is Public.

The patch makes this possible so you can just pass a client_id (without secret) to revoke a token.

@ib-lundgren could you please take a look at this one :)


josephtate:
LGTM

masci:
LGTM too

samjacoby:
Awesome! Just ran across :)

Issue #340, opened 2015-05-09
getattr broken on common.Request


kdazzle:
If I call `getattr(request, 'foo', {})`, `None` is returned instead of an empty dictionary.

This is because the `__getattr__` magic method is supposed to raise an AttributeError if the given attribute isn't found in the object ([see here](https://docs.python.org/2/reference/datamodel.html#object.__getattr__)), whereas right now it always returns `None`. 

This is undesirable because it breaks the expected functionality of `getattr` and could cause unexpected bugs.

thedrow:
This is now fixed by #340.

PR #341, opened 2015-05-09
#340 - calling getattr on Request should raise an AttributeError if that attribute doesn't exist

kdazzle:
https://github.com/idan/oauthlib/issues/340

Raise an `AttributeError` from `common.Request.__getattr__` so that `getattr` can be used like normal on a request. This commit breaks a lot of existing tests, which I'll fix in another commit.

This might break backwards compatibility if people are relying on `getattr(Request)` automatically returning `None`.

### DIFF
--- a/oauthlib/common.py
+++ b/oauthlib/common.py
@@ -381,7 +381,15 @@ class Request(object):
         self.decoded_body = extract_params(encode(body))
         self.oauth_params = []
 
-        self._params = {}
+        self._params = {
+            "client_id": None,
+            "grant_type": None,
+            "redirect_uri": None,
+            "response_type": None,
+            "state": None,
+            "refresh_token": None,
+            "access_token": None,
+        }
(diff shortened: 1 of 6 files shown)


kdazzle:
Hmm, well a _lot_ of tests fail for a lot of different places around the code. One solution would be to replace all of the assumptions that a request attribute will be either something or `None`, or another would be to provide default values in `Request._params` in the constructor.

thedrow:
I'm really not sure about this one.
Will this change break backwards compatibility?

foxx:
It's certainly the correct way of doing things, and the test appear to pass fine, although there is a chance this might be breaking something which isn't well tested, as well as other people relying on this implementation approach. I'm +1 on this, but will need to be highlighted in the change log.

kdazzle:
Thanks for taking a look at this and for merging it in!
Yeah, as @foxx said, it could break peoples' code if they are relying on that implementation, but I think this change is worth it.

PR #342, opened 2015-05-13
send no state in the access token response

g10f:
In the Authorization Code Grant flow, the state parameter is send back to the client in the authorization response.
It is useless to send the state parameter also in the access token response
because the client does a direct HTTP  request.
See https://tools.ietf.org/html/rfc6749#page-30


thedrow:
Will this break existing code?

g10f:
The change will not break any existing code.

thedrow:
Can you please rebase this?

g10f:
done

PR #350, opened 2015-06-26
Handle empty/non-parsable query strings

cheif:
We've had a issue where a trailing & on the query string 500:s the server. This PR returns a 400 instead on un-parseable query strings.


thedrow:
Thanks!

PR #355, opened 2015-07-02
Hide password in logs

lepture:
Reported by @H0neyBadger https://github.com/idan/oauthlib/issues/345

It is definitely a security bug. We shouldn't track user password in any
situation.


foxx:
LGTM.

```
>>> import re
>>> a = "username=abc&password=123&hello=world"
>>> PASSWORD_PATTERN = re.compile(r'password=[^&]+')
>>> print(PASSWORD_PATTERN.sub('password=***', a))
username=abc&password=***&hello=world
```

lepture:
There are situations that the user's password is `***`. But will we take the output of `username=abc&password=***&hello=world` as  a password exposure?

foxx:
If someone's password is ***, they deserved to get pwnt, just my two cents.

lepture:
Anyone against the patch? Anyone would +1 on it?

foxx:
Lgtm, +1

masci:
:+1:

PR #356, opened 2015-07-02
A little performance improvement

lepture:
These are some performance improvement I found when I was doing https://github.com/idan/oauthlib/pull/355


thedrow:
LGTM.

PR #358, opened 2015-07-06
Fix management of rfc6749 errors

garciasolero:
This pull request includes some fixes about management of rfc6749 errors:
    - First, check duplicate parameters
    - Missing parameter is invalid_request error according to rfc6749 
    - mismatching_redirect_uri, missing_redirect_uri,invalid_redirect_uri are not defined in rfc6749 or OAuth Extensions Error Registry
    - Check if response_type is present with a valid value according to rfc6749, then check if client is authorized

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/implicit.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/implicit.py
@@ -260,6 +260,15 @@ class ImplicitGrant(GrantTypeBase):
         # error and MUST NOT automatically redirect the user-agent to the
         # invalid redirection URI.
 
+        # First check duplicate parameters
+        for param in ('client_id', 'response_type', 'redirect_uri', 'scope', 'state'):
+            try:
+                duplicate_params = request.duplicate_params
+            except ValueError:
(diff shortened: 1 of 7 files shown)


thedrow:
Merged! Thanks.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -304,27 +313,21 @@ def validate_token_request(self, request):
...
+        # REQUIRED.
+        if request.response_type is None:
+            raise errors.MissingResponseTypeError(request=request)
+        # Value MUST be set to "token".
+        elif request.response_type != 'token':
             raise errors.UnsupportedResponseTypeError(request=request)
 
         log.debug('Validating use of response_type token for client %r (%r).',
                   request.client_id, request.client)
         if not self.request_validator.validate_response_type(request.client_id,
```
indera: Is there any point in calling  the self.request_validator.validate_response_type() after we do

indera:
It looks like the original code was allowing to extend the meaning of the "validate_response_type()" and this commit introduced a regression...

thedrow:
A regression how? Please open a new issue about it.

PR #362, opened 2015-07-20
Add token_type_hint to the list of default Request params

masci:
From v.1.0, accessing a missing field on request objects raises an AttributeError and this breaks the revocation endpoint, where an access to the `request.token_type_hint` attribute is performed during request validation.
Not sure if other parts of the library are affected, so I preferred to fix `Request`'s constructor instead of catching the error in the `RevocationEndpoint` class.

### DIFF
--- a/oauthlib/common.py
+++ b/oauthlib/common.py
@@ -397,6 +397,7 @@ class Request(object):
             "state": None,
             "token": None,
             "user": None,
+            "token_type_hint": None,
         }
         self._params.update(dict(urldecode(self.uri_query)))
         self._params.update(dict(self.decoded_body or []))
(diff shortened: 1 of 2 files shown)


masci:
@lepture, @tilgovi, @georgewhewell, @samjacoby, @foxx, @thedrow anyone available for a quick review of this PR?

lepture:
Yes token_type_hint  is optional: https://tools.ietf.org/html/rfc7009#section-2.1

masci:
@lepture do you think this is enough to draft a 1.0.1? I really need this fix in Django OAuth Toolkit.

thedrow:
@masci I can release 1.0.1 if needed.

PR #366, opened 2015-07-23
Fix revoke process when working with django-oauth-toolkit

abarto:
A fix revocation process when working with django-oauth-toolkit using safer attribute access of request's fields on revocation. There might be an underlying issue (maybe the referenced field was changed?), but this quick fix does the trick.


lepture:
This is duplicated with https://github.com/idan/oauthlib/pull/362

thedrow:
Thanks for your contribution anyway.

PR #371, opened 2015-08-04
Python3 fix: encodes request body before hashing

mitgr81:
In python 3 hashlib works on bytestrings, so we need to encode arguments
to most, if not all, hashing functions.


thedrow:
Thanks for the contribution.

PR #381, opened 2015-08-12
adding validator_log to store what the endpoint computed

jvanasco:
This adds logging of what the endpoint computed to the request as a dict on the  `validator_log` attribute.  

see https://github.com/idan/oauthlib/issues/378

This is most useful on protected resources, where it lets you know why the overall validation failed.


thedrow:
Thanks for the contribution.

PR #391, opened 2015-09-04
Fix for null scope

vrehak:
Our PHP OAuth2 server responds to token request with
{"access_token":"20fdab98f6518335db837c2039125ae16b523eb86","expires_in":"7200","token_type":"Bearer","scope":null}

Without this fix, parse_token_response fails with:
  File "/usr/local/lib/python2.7/dist-packages/oauthlib/oauth2/rfc6749/tokens.py", line 32, in **init**
    self._new_scope = set(utils.scope_to_list(params['scope']))
TypeError: 'NoneType' object is not iterable

The reason is that scope_to_list returns None for this case and set(None) results in TypeError. It feels logical to me that scope_to_list should always return a list.

### DIFF
--- a/oauthlib/oauth2/rfc6749/utils.py
+++ b/oauthlib/oauth2/rfc6749/utils.py
@@ -35,7 +35,7 @@ def scope_to_list(scope):
     if isinstance(scope, (tuple, list, set)):
         return [unicode_type(s) for s in scope]
     elif scope is None:
-        return None
+        return []
     else:
         return scope.strip().split(" ")
 


thedrow:
I'll have to check if this was done for a reason. I'll get back to you.

lig:
Hello! I have exactly the same issue with PHP OAuth2 provider used in SugarCRM.

thedrow:
I think we uncovered a different bug here.
> If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined).
The scope can't be empty. If the default scope is empty the server should report an error.
See http://tools.ietf.org/html/rfc6749#section-3.3

lig:
@thedrow the quest is does `null` stands for "empty"?

thedrow:
@lig Yes, I believe it is.

thedrow:
Right we need to provide a default scope. I'll open an issue about it.

PR #394, opened 2015-09-21
Add 'python' info to README

cyrilchaponeverysens:
Updated readme to say that the library is for *Python*.

I landed here just searching for generic **oauth** docs, and it took me like 30 seconds to realize that this library is a Python-specific implementation.


thedrow:
Thanks!

PR #407, opened 2016-03-11
Allow `/` and `?` characters in query string (RFC 3986 section 3.4)

eofs:
Current implementation doesn't allow `/` and `?` character to appear in query string in non-encoded form.

However RFC 3986 section 3.4 states following:

> However, as query components are often used to carry identifying information in the form of "key=value" pairs and one frequently used value is a reference to another URI, it is sometimes better for usability to avoid percent-encoding those characters.

This pull requests adds `/` and `?` characters into allowed characters list.


thedrow:
The tests fail on 3.2 because virtualenv dropped support. Disregard that.
I think there was a reason why these were disallowed though I'm not sure.

thedrow:
No you're right.

PR #416, opened 2016-04-20
Openid connect

joelstevenson:
Rebasing the work done by @ib-lundgren in 2014 onto the current version of master and providing some additional work to get OpenID Connect generally useable.  Authorization Code flow is the primary target but existing work on Implicit and Hybrid is still included.

### DIFF
--- a/oauthlib/oauth2/rfc6749/endpoints/authorization.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/authorization.py
@@ -64,11 +64,7 @@ class AuthorizationEndpoint(BaseEndpoint):
     def __init__(self, default_response_type, default_token_type,
                  response_types):
         BaseEndpoint.__init__(self)
-        self._response_types = {}
-        # response_types are sorted internally so ordered comparison is faster/easier later
-
-        for k, v in response_types.iteritems() if sys.version_info[0] == 2 else iter(response_types.items()):
-            self._response_types[",".join(sorted(k.split()))] = v
+        self._response_types = response_types
 
         self._default_response_type = default_response_type
         self._default_token_type = default_token_type
@@ -89,16 +85,6 @@ class AuthorizationEndpoint(BaseEndpoint):
     def default_token_type(self):
         return self._default_token_type
 
-    def get_response_types_handler(self, request):
-
-        if request.response_type is None:
-            response_type_handler = self.default_response_type_handler
-        else:
-            response_type_handler = self.response_types.get(
-                ",".join(sorted(request.response_type.split())), self.default_response_type_handler)
-
-        return response_type_handler
-
     @catch_errors_and_unavailability
     def create_authorization_response(self, uri, http_method='GET', body=None,
                                       headers=None, scopes=None, credentials=None):
@@ -110,7 +96,8 @@ class AuthorizationEndpoint(BaseEndpoint):
         request.user = None     # TODO: explain this in docs
         for k, v in (credentials or {}).items():
             setattr(request, k, v)
-        response_type_handler = self.get_response_types_handler(request)
+        response_type_handler = self.response_types.get(
+            request.response_type, self.default_response_type_handler)
         log.debug('Dispatching response_type %s request to %r.',
                   request.response_type, response_type_handler)
         return response_type_handler.create_authorization_response(
@@ -123,5 +110,6 @@ class AuthorizationEndpoint(BaseEndpoint):
         request = Request(
             uri, http_method=http_method, body=body, headers=headers)
         request.scopes = None
-        response_type_handler = self.get_response_types_handler(request)
+        response_type_handler = self.response_types.get(
+            request.response_type, self.default_response_type_handler)
         return response_type_handler.validate_authorization_request(request)
--- a/oauthlib/oauth2/rfc6749/endpoints/authorization.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/authorization.py
@@ -7,7 +7,6 @@ This module is an implementation of various logic needed
 for consuming and providing OAuth 2.0 RFC6749.
 """
 from __future__ import absolute_import, unicode_literals
-import sys
 import logging
 
 from oauthlib.common import Request
@@ -65,7 +64,6 @@ class AuthorizationEndpoint(BaseEndpoint):
                  response_types):
         BaseEndpoint.__init__(self)
         self._response_types = response_types
-
         self._default_response_type = default_response_type
         self._default_token_type = default_token_type
 
--- a/oauthlib/oauth2/rfc6749/endpoints/resource.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/resource.py
@@ -9,8 +9,10 @@ for consuming and providing OAuth 2.0 RFC6749.
 from __future__ import absolute_import, unicode_literals
 
 import logging
+import json
 
 from oauthlib.common import Request
+from ..errors import InvalidRequestError
 
 from .base import BaseEndpoint, catch_errors_and_unavailability
 
@@ -72,7 +74,20 @@ class ResourceEndpoint(BaseEndpoint):
                                              self.default_token_type_handler)
         log.debug('Dispatching token_type %s request to %r.',
                   request.token_type, token_type_handler)
-        return token_type_handler.validate_request(request), request
+        is_valid = token_type_handler.validate_request(request)
+
+        # the validate_request() call is a chance for the client library to finalize the scopes on the request
+        if is_valid:
+            if request.claims and request.scopes and "openid" in request.scopes:
+                # specific claims are requested, i.e. as part of a UserInfo endpoint request
+                # see http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
+                try:
+                    request.claims = json.loads(request.claims)
+                except:
+                    raise InvalidRequestError(description="Malformed claims parameter",
+                                              uri="http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter")
+
+        return is_valid, request
 
     def find_token_type(self, request):
         """Token type identification.
--- a/oauthlib/oauth2/rfc6749/endpoints/resource.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/resource.py
@@ -9,10 +9,8 @@ for consuming and providing OAuth 2.0 RFC6749.
 from __future__ import absolute_import, unicode_literals
 
 import logging
-import json
 
 from oauthlib.common import Request
-from ..errors import InvalidRequestError
 
 from .base import BaseEndpoint, catch_errors_and_unavailability
 
@@ -74,19 +72,7 @@ class ResourceEndpoint(BaseEndpoint):
                                              self.default_token_type_handler)
         log.debug('Dispatching token_type %s request to %r.',
                   request.token_type, token_type_handler)
-        is_valid = token_type_handler.validate_request(request)
-
-        # the validate_request() call is a chance for the client library to finalize the scopes on the request
-        if is_valid and request.claims and request.scopes and "openid" in request.scopes:
-                # specific claims are requested, i.e. as part of a UserInfo endpoint request
-                # see http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
-                try:
-                    request.claims = json.loads(request.claims)
-                except:
-                    raise InvalidRequestError(description="Malformed claims parameter",
-                                              uri="http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter")
-
-        return is_valid, request
+        return token_type_handler.validate_request(request), request
 
     def find_token_type(self, request):
         """Token type identification.
--- a/oauthlib/oauth2/rfc6749/endpoints/resource.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/resource.py
@@ -77,8 +77,7 @@ class ResourceEndpoint(BaseEndpoint):
         is_valid = token_type_handler.validate_request(request)
 
         # the validate_request() call is a chance for the client library to finalize the scopes on the request
-        if is_valid:
-            if request.claims and request.scopes and "openid" in request.scopes:
+        if is_valid and request.claims and request.scopes and "openid" in request.scopes:
                 # specific claims are requested, i.e. as part of a UserInfo endpoint request
                 # see http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
                 try:
--- a/oauthlib/oauth2/rfc6749/endpoints/revocation.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/revocation.py
@@ -126,5 +126,5 @@ class RevocationEndpoint(BaseEndpoint):
 
         if (request.token_type_hint and
                 request.token_type_hint in self.valid_token_types and
-                request.token_type_hint not in self.supported_token_types):
+                not request.token_type_hint in self.supported_token_types):
             raise UnsupportedTokenTypeError(request=request)
--- a/oauthlib/oauth2/rfc6749/endpoints/revocation.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/revocation.py
@@ -126,5 +126,5 @@ class RevocationEndpoint(BaseEndpoint):
 
         if (request.token_type_hint and
                 request.token_type_hint in self.valid_token_types and
-                not request.token_type_hint in self.supported_token_types):
+                request.token_type_hint not in self.supported_token_types):
             raise UnsupportedTokenTypeError(request=request)
--- a/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
@@ -100,7 +100,7 @@ class AuthorizationCodeGrant(GrantTypeBase):
     def __init__(self, request_validator=None, refresh_token=True):
         self.request_validator = request_validator or RequestValidator()
         self.refresh_token = refresh_token
-        # NEW-FOR-OPENID
+
         self._authorization_validators = []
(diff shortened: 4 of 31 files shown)


tox.ini
```diff
@@ -1,10 +1,12 @@
 [tox]
-envlist = py26,py27,py32,py33,py34,py35,pypy,pypy3
```
clintonb: Why the changes to this file?

joelstevenson: Woops, I didn't mean to include that.  End of day mistake.

tox.ini
```diff
@@ -1,10 +1,12 @@
...
+;envlist = py26,py27,py32,py33,py34,py35,pypy,pypy3
+envlist = py27
 
 [testenv]
 deps=
     -rrequirements-test.txt
-commands=nosetests --with-coverage --cover-erase --cover-package=oauthlib -w tests
+; commands=nosetests --with-coverage --cover-erase --cover-package=oauthlib -w tests tests.oauth2.rfc6749.grant_types.test_openid_connect
```
thedrow: Why is this comment here?

thedrow:
Nice job. Have you tested this yourself?

oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
```diff
@@ -95,10 +95,35 @@ class AuthorizationCodeGrant(GrantTypeBase):
...
+    def register_token_validator(self, validator):
+        self._token_validators.append(validator)
+
+    def register_code_modifier(self, modifier):
+        self._code_modifiers.append(modifier)
+
+    def register_token_modifier(self, modifier):
+        self._token_modifiers.append(modifier)
+
+    # END-NEW-FOR-OPENID
```
bjmc: Should this have a `#Start of OpenIDConnect` section? Also, maybe rephrase this because it's new now, but it won't always be new.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -319,7 +335,7 @@ def validate_token_request(self, request):
         if request.response_type is None:
             raise errors.MissingResponseTypeError(request=request)
         # Value MUST be set to "token".
-        elif request.response_type != 'token':
+        elif not request.response_type in self.response_types:
```
bjmc: IIRC, the [OpenIDConnect spec allows](https://openid.net/specs/openid-connect-core-1_0.html#Authentication) specifying multiple, space-separated response types.

Should this line be something like:

`if not set(self.request.response_type.split()).issubset(self.response_types):`

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -0,0 +1,408 @@
...
+        if request.max_age:
+            d = datetime.datetime.utcnow()
+            token['auth_time'] = d.isoformat("T") + "Z"
+
+        # TODO: acr claims (probably better handled by server code using oauthlib in get_id_token)
+
+        token['id_token'] = self.request_validator.get_id_token(token, token_handler, request)
+
+        if request.response_type == "id_token":
+            # implicit flow, only requesting id_token so don't return the access token
```
bjmc: This feels a bit hacky. If I understand correctly, in this scenario, we've already generated an access token, but then we're just throwing it away and not returning it to the user? I guess that's fine, but it seems like we should be able to do better and avoid creating access tokens we don't need. Can we dispatch on response_type earlier in the flow somewhere?

oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -15,13 +16,22 @@
 class GrantTypeBase(object):
     error_uri = None
     request_validator = None
+    default_response_mode = 'fragment'
```
bjmc: Why is this the default on the GrantTypeBase?

oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -15,13 +16,22 @@
...
 
     def create_authorization_response(self, request, token_handler):
         raise NotImplementedError('Subclasses must implement this method.')
 
     def create_token_response(self, request, token_handler):
         raise NotImplementedError('Subclasses must implement this method.')
 
+    def add_token(self, token, token_handler, request):
+        # Only add a hybrid access token on auth step if asked for
+        if not 'token' in request.response_type.split():
```
bjmc: Can we modify the request object to do this split() for us and have `request.response_type` always be a set? Or will that break things elsewhere?

tox.ini
joelstevenson: Shouldn't be.  I'll pull that out too.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
joelstevenson: Ah yes quite possibly.  I'll review this.

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
joelstevenson: Quite possibly, I'll look into what it will take.

oauthlib/oauth2/rfc6749/grant_types/base.py
joelstevenson: It looks like this was added to satisfy the default fragment response type in the OpenID spec, but I think this partially conflicts with the defaults defined in the OAuth2 Multiple Response Modes spec.  It may not impact existing OAuth2 flows (at least didn't reflect as an error in the tests) but I'll review to see if this can be less far reaching and scoped more toward OpenID specific requests.

oauthlib/oauth2/rfc6749/grant_types/base.py
joelstevenson: Sounds like a good thing to do, at least during OpenID specific requests.  I'll review.

joelstevenson:
@thedrow I'm developing it in the context of a Django app using Django-OAuth2-Toolkit, with a small few patches to DOT to support any changes made here to the oauthlib (mostly the validator) .  So far so good.

thedrow:
Ok let me know when you guys are ready.

oauthlib/oauth2/rfc6749/grant_types/base.py
joelstevenson: I think splitting `request.response_type` will impact server implementations by basically changing it out from under them.  We could split and cache it in a private property like `_request_type` but I don't know that it's worth it here.  Thoughts?

oauthlib/oauth2/rfc6749/grant_types/base.py
joelstevenson: `response_mode` defaults to "fragment" for all grants other than AuthorizationCode.  That grant overrides it and Implicit and Hybrid don't so it's mostly DRY.

oauthlib/oauth2/rfc6749/grant_types/base.py
bjmc: I think splitting at all might be the wrong behavior: After reading [this document](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#Combinations) referenced from the OIDC spec, it sounds like we shouldn't be treating the response type as a space-separated list of values, but rather an enum that can take specific values of: `code token`, `code id_token`, `id_token token` or `code id_token token`.

joelstevenson: True, I guess it's a question of how rigid we want to be about ordering of values in the `response_type`: should we recognize `token code` and `code token` as the same or only accept it in the `code token` form?  I'm inclined to be accepting of both given that there aren't semantic differences between them, but I can see an argument being made for cleaving to the letter of the spec too.

bjmc: I agree it's instinctively appealing to accept both `token code` and `code token`, but I wonder if that could lead to issues down the road, or conflicts with future versions of the spec?

How do other people feel? @clintonb @thedrow ?

oauthlib/oauth2/rfc6749/endpoints/authorization.py
```diff
@@ -64,7 +64,12 @@ class AuthorizationEndpoint(BaseEndpoint):
     def __init__(self, default_response_type, default_token_type,
                  response_types):
         BaseEndpoint.__init__(self)
-        self._response_types = response_types
+        self._response_types = {}
+        # response_types are sorted internally so ordered comparison is faster/easier later
+
+        for k, v in response_types.iteritems() if sys.version_info[0] == 2 else iter(response_types.items()):
```
bjmc: This reads really ugly. Can we just use `response_types.items()` for both Python 2 and Python 3? I know it's slightly less memory-efficient in Py2, but `response_types` is going to be pretty darn small, so I doubt it matters.

Failing that, I think we should use [`six`](https://pythonhosted.org/six/) to support multi-version Python, instead of scattering compatibility code throughout our code base.

oauthlib/oauth2/rfc6749/endpoints/authorization.py
```diff
@@ -64,7 +64,12 @@ class AuthorizationEndpoint(BaseEndpoint):
...
+            self._response_types[",".join(sorted(k.split()))] = v
```
bjmc: Could we use a `FrozenSet` instead of a comma-separated string for this purpose?

oauthlib/oauth2/rfc6749/request_validator.py
```diff
@@ -270,6 +277,20 @@ def save_bearer_token(self, token, request, *args, **kwargs):
         """
         raise NotImplementedError('Subclasses must implement this method.')
 
+    def get_id_token(self, token, token_handler, request):
```
bjmc: Should this take a `nonce` explicitly? Or is it expected that implementors will get the nonce off the request object? If so, it might be good to document that.

oauthlib/oauth2/rfc6749/endpoints/resource.py
```diff
@@ -72,7 +74,19 @@ def verify_request(self, uri, http_method='GET', body=None, headers=None,
                                              self.default_token_type_handler)
         log.debug('Dispatching token_type %s request to %r.',
                   request.token_type, token_type_handler)
-        return token_type_handler.validate_request(request), request
+        is_valid = token_type_handler.validate_request(request)
+
+        # the validate_request() call is a chance for the client library to finalize the scopes on the request
```
bjmc: :+1: Supporting the `claims` parameter.

Maybe I'm not understanding, but are we sure this should be done as part of `token_type_handler.validate_request()` or should it be another method on the RequestValidator?

joelstevenson: Yeah good question.  Currently the `validate_request()` method calls through to the request_validator's  `validate_bearer_token()` method where the validator currently has the chance to load the token from whatever persistent store it is using and populate request attributes from the stored token's data.  Both flask_oauthlib and the django-oauth-toolkit are doing this and setting the `client`, `user`, and `access_token` properties of the `request` (my patched version of DOT also sets `scopes` from the database because they may not come in via the HTTP request).  Assuming other clients are taking the same approach I didn't want to introduce new interface methods for clients to implement without a strong reason too.

Now that I've been poking around in DOT and oauthlib for a while I can envision various interface level changes but I've been trying to minimize the impact on the RequestValidator interface definition to keep the update path clear.

oauthlib/oauth2/rfc6749/request_validator.py
joelstevenson: In general it appears that the expectation has been that clients will be pulling any relevant parameters off of the `request`.  I'll make a note in the docs about the nonce in implicit and hybrid workflows.

oauthlib/oauth2/rfc6749/endpoints/authorization.py
joelstevenson: Fair points.  On further rumination I think supporting randomly ordered response_type values is not a significant use case and the code will be cleaner overall if we simply look for `response_type` values as defined in the spec's registry.  Code reverting this is pushed.

bjmc: I think that's probably the right decision. Nobody can really complain at us for following the letter of the spec.

oauthlib/oauth2/rfc6749/endpoints/resource.py
thedrow: If there will be any API breakage I'll release this as oauthlib 2.0 and libraries will upgrade when they see fit.

bjmc: I'll also note that we could probably safely add methods to RequestValidator as long as the RequestValidator base class contains no-op stubs of the relevant methods. Presumably implementers would be subclassing the Base, and so they wouldn't have any breakage.

joelstevenson: Reviewing the flow here it looks like my comment 

> the validate_request() call is a chance for the client library to finalize the scopes on the request 

there is wrong.  `request.scopes` ought to be the scopes _required to access_ the Resource and can be defined by the client library prior to calling our `verify_request()`.  It's a bit confusing - if anything I'd rename the `scopes` argument of `verify_request()` to be `required_scopes`.  That would only slightly clarify it though - the `request.scopes` property is overloaded to mean both "scopes listed in the incoming HTTP request" in other request flows like Authorization and Token requests and "scopes required to access the requested resource" as in this protected resource request flow.

joelstevenson: I'm ok with this living in `validate_request()` given the way the `request` is currently being populated within that code flow but if there's a clear desire to move it into a new `RequestValidator` method where this `claims` parameter parsing takes place let me know.

joelstevenson:
@thedrow @bjmc @clintonb @ib-lundgren I think this is pretty much ready - but does anyone have any lingering concerns or suggestions?

thedrow:
How much API does this PR break if at all?

thedrow:
Also, this feature really needs to be documented.

joelstevenson:
Sorry for the delay - I hope to get to that this week.

clintonb:
@joelstevenson @thedrow @ib-lundgren is this still under active development? What can we do to get it delivered?

joelstevenson:
@clintonb I need to provide some documentation describing what aspects of OIDC are supported.  Thanks for the ping - I will do so today.

joelstevenson:
@thedrow the OpenID Connect auth flows don't break any existing API, as such, though the do extend it when an OpenID Connect auth request is received.  The OIDC code is only active when/if the request has a scope including `openid`.  For all other requests OAuthLib will behave as if unchanged.  

If OAuthLib users are passing the `scope` parameter through unfiltered (which is likely I'd imagine), and a request comes through that triggers the OIDC code then it will result in a `NotImplementedError` from one of:

```
RequestValidator.validate_silent_authorization()
RequestValidator.validate_silent_login()
RequestValidator.validate_user_match()
RequestValidator.get_id_token()
```

Until their validator subclass provides an implementation.

bjmc:
Is there anything else needed to push this across the finish line?

oauthlib/oauth2/rfc6749/endpoints/revocation.py
```diff
@@ -126,5 +126,5 @@ def validate_revocation_request(self, request):
 
         if (request.token_type_hint and
                 request.token_type_hint in self.valid_token_types and
-                request.token_type_hint not in self.supported_token_types):
+                not request.token_type_hint in self.supported_token_types):
```
thedrow: Why use `not var in list` instead of `var not in list`?

oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -39,3 +49,38 @@ def validate_scopes(self, request):
...
+
+    def prepare_authorization_response(self, request, token, headers, body, status):
+        """Place token according to response mode.
+
+        Base classes can define a default response mode for their authorization
+        response by overriding the static `default_response_mode` member.
+        """
+        request.response_mode = request.response_mode or self.default_response_mode
+
+        if not request.response_mode in ('query', 'fragment'):
```
thedrow: Here `not var in tuple` appears again.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -319,7 +342,7 @@ def validate_token_request(self, request):
         if request.response_type is None:
             raise errors.MissingResponseTypeError(request=request)
         # Value MUST be set to "token".
```
thedrow: Is the comment above still correct?

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -0,0 +1,442 @@
...
+                raise InvalidRequestError(description="Malformed claims parameter",
+                                          uri="http://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter")
+
+    def add_id_token(self, token, token_handler, request):
+        # Treat it as normal OAuth 2 auth code request if openid is not present
+        if not 'openid' in request.scopes:
+            return token
+
+        # Only add an id token on auth/token step if asked for.
+        if request.response_type and not 'id_token' in request.response_type:
```
thedrow: Again `not var in list`...

thedrow:
@bjmc Just a few more minor issues.
@joelstevenson Can you take a look at the comments?
Also, you can add yourself to the AUTHORS file :)

thedrow:
@joelstevenson Great job! Thank you for your contribution.
I'll release this feature tonight (Israel Time) :)

joelstevenson:
@thedrow .  Cheers, I'm happy to help!  Most of the work was from @ib-lundgren so a big thanks to him for bootstrapping the whole effort.

thedrow:
Released!

PR #423, opened 2016-05-09
Fix #422: Query strings should be able to include colons

### LINKED ISSUE #422: Query strings should be able to include colons: `:`
Can we update https://github.com/idan/oauthlib/blob/master/oauthlib/common.py#L112 to include colons as well?

r4fek:


thedrow:
Is there an RFC somewhere I can look at? I want to resolve these types of issues once and for all.

PR #425, opened 2016-05-16
Cast body to a string for regex - fixes #424

### LINKED ISSUE #424: Latest change to __repr__ breaks Flask-Oauthlib?
Hi,
The regex substitution on the body introduced in the latest PR breaks functionality for Flask-Oauthlib.
There they pass a body as a dictionary for validation, so regex can't work.

Should the body be cast to string before sanitization?

parryjacob:
This just ensures that when we go to sanitize the body for logging, it is actually a string

### DIFF
--- a/oauthlib/common.py
+++ b/oauthlib/common.py
@@ -416,7 +416,7 @@ class Request(object):
         body = self.body
         headers = self.headers.copy()
         if body:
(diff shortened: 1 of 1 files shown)


thedrow:
Thanks. Will release soon. In the meanwhile you can use master.

thedrow:
Released as 1.1.2. It will be available on PyPi soon.

Issue #45, opened 2012-07-05
OAuth2 Server Discussion


emberian:
@ib-lundgren

I need an OAuth2 authorization server as well as a resource server, and no existing implementations exist besides some django-specific stuff. oauthlib looks like a worthy place to contribute one. I'm willing to do all of the work, as I need this to do my job, but I'd like some discussion. My initial thoughts after reading the spec:
- There is leeway for lots of extensibility, which might be difficult to wrap
- There isn't actually a lot that needs to be done

Similar to the discussion of the client, the only thing that really needs to be done here is provide a nice API. I think an interface very similar to oauthlib.oauth1.rfc5849.Server would be fine.

PR #437, opened 2016-09-17
Fix TypeError in openid_connect

### LINKED ISSUE #436: TypeError: argument of type 'NoneType' is not iterable in openid_connect implementation
When trying to use the OAuth2 authorization endpoint (indirectly from a django-oauth-toolkit based application) I found an issue with how oauthlib/oauth2/rfc6749/endpoints/authorization.py and oauthlib/oauth2/rfc6749/grant_types/openid_connect.py interact. I assume that the problem is not related to django-oauth-toolkit because I can easily reproduce it with a minimal test in an interactive Python interpreter:

```
Python 2.7.11+ (default, May  9 2016, 15:54:33) 
[GCC 5.3.1 20160429] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from oauthlib.oauth2.rfc6749.endpoints import pre_configured
>>> ps=pre_configured.Server(None)
>>> ps.validate_authorization_request('/oauth2/authorize/', 'http://localhost:8080/', body='', headers={})
  File "<stdin>", line 1, in <module>
  File "/home/jan/.virtualenvs/test/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/endpoints/base.py", line 64, in wrapper
    return f(endpoint, uri, *args, **kwargs)
  File "/home/jan/.virtualenvs/test/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/endpoints/authorization.py", line 113, in validate_authorization_request
    return response_type_handler.validate_authorization_request(request)
  File "/home/jan/.virtualenvs/test/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py", line 77, in validate_authorization_request
    return self._handler_for_request(request).validate_authorization_request(request)
  File "/home/jan/.virtualenvs/test/local/lib/python2.7/site-packages/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py", line 67, in _handler_for_request
    if "openid" in request.scopes:
TypeError: argument of type 'NoneType' is not iterable
```

The validate_authorization_request method in authorization.py initializes request.scope:

``` python

request.scopes = None
```

and the _handler_for_request method in openid_connect.py tries to iterate request.scopes

``` python

if "openid" in request.scopes:
    handler = self.oidc_auth_grant
```

I walked through the code in pdb but there is nothing setting the request.scope to any useful value between these calls. I tried to set request.scopes to [] in validate_authorization_request which fixes the TypeError but I do not know whether this is semantically correct.

I used oauthlib 2.0.0, but the code is the same as in the current master branch. I don't think that this is a duplicate of #433 or #359.

jandd:
This pull request fixes the "TypeError: argument of type 'NoneType' is not iterable" that was caused by calling the openid_connect validate_authorization method with a request where request.scopes was
set to None (fixes #436).

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
@@ -64,7 +64,7 @@ class AuthCodeGrantDispatcher(object):
     def _handler_for_request(self, request):
         handler = self.default_auth_grant
 
-        if "openid" in request.scopes:
+        if request.scopes and "openid" in request.scopes:
             handler = self.oidc_auth_grant
 
         log.debug('Selecting handler for request %r.', handler)


jandd:
The tox run for py32 in the Travis build fails for reasons not related to the code change.

jandd:
The build fails for the cloned master branch as well (see https://travis-ci.org/jandd/oauthlib/jobs/160642978)

yon2004:
Can confirm that this fixes the issue. Thanks @jandd

Sumukh:
If you merge master back in - tests should pass again. 

I'm not sure who is responsible for merging/reviewing - but the change in here is really small and can only fix errors (which is impacting a lot of users of downstream libraries).

thedrow:
#45 Should fix this

PR #441, opened 2016-10-17
Updating request.scopes to a list to fix the NoneType is not iterable error.

davidvartanian:
Scopes are initialised as request.scopes = None in validate_authorization_request() method and response_type_handler.validate_authorization_request(request) is called immediately afterwards.
Now, in method _handler_for_request() called by validate_authorization_request() there's a call to if "openid" in request.scopes, which expects of course request.scopes to be an Iterable.

Since I couldn't find any other way this library to work, please merge it so that it's available to be installed via pip.
Thanks!


davidvartanian:
Please notice that Travis tests fail on py32 for reasons not related to this commit as well as the tests on master without my change.

Sumukh:
#436 is also another approach to this same issue. What we've done is install directly from the git repo. 

I have a fork that should work:  https://github.com/Sumukh/oauthlib

thedrow:
#450 should fix this

PR #442, opened 2016-10-21
scrubbing rsa_key from __repr__

mistermocha:
self.rsa_key is printing out another secret! This should be scrubbed from **repr**


thedrow:
Thanks!

PR #449, opened 2016-11-21
Improve validator registration

bjmc:
This changeset is an attempt to improve and standardize the validator registration hooks introduced with the OpenIDConnect changes.

- Moves custom validator registration up onto the `GrantTypeBase` class and removes those methods from the subclass grants.
- DRYs up the OpenIDConnect grant type classes to be more of thin proxies onto the standard OAuth2 grant classes.
- Adds the ability to register validators that run either before or after the standard validation methods.
- Adds custom validator invocations to `ClientCredentialsGrant`, `RefreshTokenGrant` and `ResourceOwnerPasswordCredentialsGrant`.
- Adds tests for custom validators.

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/base.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/base.py
@@ -4,6 +4,7 @@ oauthlib.oauth2.rfc6749.grant_types
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 """
 from __future__ import unicode_literals, absolute_import
+from itertools import chain
 
 import logging
 
@@ -12,6 +13,28 @@ from oauthlib.oauth2.rfc6749 import errors, utils
 from ..request_validator import RequestValidator
 
 log = logging.getLogger(__name__)
+
+class ValidatorsContainer(object):
+
+    """
+        Container object for holding validator callables to be invoked.
+    """
+
+    def __init__(self,
+                 post_auth=None, post_token=None,
+                 pre_auth=None, pre_token=None):
+        self.pre_auth = pre_auth
+        self.post_auth = post_auth
+        self.pre_token = pre_token
+        self.post_token = post_token
+
+    @property
+    def all_pre(self):
+        return chain(self.pre_auth, self.pre_token)
+
+    @property
+    def all_post(self):
+        return chain(self.post_auth, self.post_token)
 
 
 class GrantTypeBase(object):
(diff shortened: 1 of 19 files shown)


thedrow:
This looks great. Can you add some documentation?
If you can't just let me know and I'll pick it up as soon as I'll be able to.

thedrow:
@bjmc Ping?

bjmc:
Yeah, I can add some documentation. Do you just want docstrings? Or as part of a separate document?

thedrow:
Docsstrings inline and usage documentation under `/docs` is possible. Thanks.
If you are not already on the AUTHORS file feel free to add yourself.

bjmc:
@thedrow I hope that looks okay for the docs, I'm not an expert at the sphinx syntax.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -328,6 +317,16 @@ def validate_token_request(self, request):
 
         # Then check for normal errors.
 
+        request_info = {}
+        # For implicit grant, auth_validators and token_validators are
+        # basically equivalent since the token is returned from the
+        # authorization endpoint.
+        for validator in chain(self._token_validators_run_before_standard_ones,
```
thedrow: Extract to a helper method since this code is repeated on line 372 with different validators.

oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -17,6 +18,75 @@ class GrantTypeBase(object):
...
+                                    validations. If False, the custom valdator is called before
+                                    the standard validations.
+        :returns: None
+        """
+        if after_standard:
+            self._auth_validators_run_after_standard_ones.append(validator)
+        else:
+            self._auth_validators_run_before_standard_ones.append(validator)
+
+    def register_token_validator(self, validator, after_standard=True):
```
bjmc: I really went back and forth on this. I agree the boolean is a little bit clunky.

Other options:

**Use two separate methods (four when you count auth validators and token validators).**

The reasons I didn't do this were:

1) Using the boolean let me preserve backwards compatibility for anyone who is already using `register_*_validator()` in its existing incarnation.

2) Many users just want some custom validations in the mix, and don't care about whether they run before or after the existing ones. Having a boolean with a default spares those users from having to think about it.

3) I couldn't think of method names that were informative without being so verbose it seemed unwieldy, e.g. `register_auth_validator_run_before_standard()` (before standard what?)

**Just completely get rid of these registration methods and let callers append or insert to the list of validators directly.**

I actually kind of like this option, because it feels more Pythonic doing away with setter methods, but I didn't do it because it seemed out of step with the rest of the code base, and obviously it somewhat locks you into an implementation for the future.

oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -17,6 +18,75 @@ class GrantTypeBase(object):
     error_uri = None
     request_validator = None
     default_response_mode = 'fragment'
+    refresh_token = True
+    response_types = ['code']
+
+    def __init__(self, request_validator=None, **kwargs):
+        self.request_validator = request_validator or RequestValidator()
```
bjmc: Well, you could pass `_auth_validators_run_before_standard_ones` as a keyword argument to the constructor, and that would do it. But that feels a bit hacky.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -366,13 +363,23 @@ def validate_token_request(self, request):
...
         # authorization endpoint.
-        for validator in chain(self._auth_validators_run_after_standard_ones,
-                               self._token_validators_run_after_standard_ones):
+        for validator in chain(auth_validators, token_validators):
             result = validator(request)
             if result is not None:
                 request_info.update(result)
-
-        return request.scopes, request_info
+        return request_info
```
kevin-brown: This method returns `request_info`, but this return value is not used in any of the places where it is called.

thedrow: I agree it's better.

oauthlib/oauth2/rfc6749/grant_types/base.py
thedrow: Another option would be to wrap the validators list with an API that allows you to switch the internal implementation later on.
What do you think?

bjmc: That's true. I would maybe make the case that the API we expose should be the standard Python [`MutableSequence`](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSequence) and we can start by just using a stock list. If in the future, anyone wants to change the implementation but preserve the API, they can use a `MutableSequence` subclass.

bjmc:
@thedrow Do you like that better? I moved all the custom validators into a container class, and you can now pass keyword args to the grant type constructor to initialize a `GrantType` with custom validations, instead of appending them after the fact.

thedrow:
LGTM! Thanks.

PR #450, opened 2016-11-21
Normalize handling of request.scopes list

### LINKED ISSUE #436: TypeError: argument of type 'NoneType' is not iterable in openid_connect implementation

joelstevenson:
Use the scope_to_list() util to initalize the request.scopes list from
the request.scope request parameter in two place where it was instead
being set to None.

- AuthorizationEndpoint.validate_authorization_request()
- TokenEndpoint.create_token_response()

In both cases the Request should be properly populated before it is
passed to the client's validator.

In the case of the TokenEndpoint - there are OAuth2 workflows that
allow an optional scope parameter so we should have been doing this
for them anyway.

Since scope_to_list() may return None, also update the openid_connect
code to behave properly when this is the case.

Fixes #436 cf. #437 , #441

### DIFF
--- a/oauthlib/oauth2/rfc6749/endpoints/authorization.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/authorization.py
@@ -10,6 +10,7 @@ from __future__ import absolute_import, unicode_literals
 import logging
 
 from oauthlib.common import Request
+from oauthlib.oauth2.rfc6749 import utils
 
 from .base import BaseEndpoint, catch_errors_and_unavailability
 
@@ -107,7 +108,9 @@ class AuthorizationEndpoint(BaseEndpoint):
         """Extract response_type and route to the designated handler."""
         request = Request(
             uri, http_method=http_method, body=body, headers=headers)
-        request.scopes = None
+
+        request.scopes = utils.scope_to_list(request.scope)
+
         response_type_handler = self.response_types.get(
             request.response_type, self.default_response_type_handler)
         return response_type_handler.validate_authorization_request(request)
--- a/oauthlib/oauth2/rfc6749/endpoints/token.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/token.py
@@ -11,6 +11,7 @@ from __future__ import absolute_import, unicode_literals
 import logging
 
 from oauthlib.common import Request
+from oauthlib.oauth2.rfc6749 import utils
 
 from .base import BaseEndpoint, catch_errors_and_unavailability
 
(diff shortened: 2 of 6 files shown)


thedrow:
Care to fix the test?

thedrow:
Thank you very much!

PR #453, opened 2016-12-07
Public clients should be validated when revoking tokens

kippandrew:
According to [RC7009 Section 2.1](https://tools.ietf.org/html/rfc7009#section-2.1), a client should include authentication credentials when revoking its tokens. As discussed in #339, this is not make sense for public clients. However, in that case, the public client should still be checked that is infact a public client (`authenticate_client_id`).

I also fixed a what appears to be a bad merge (`test_revoke_token_without_client_authentication` was defined twice), which is why the diff is a little messy.


kippandrew:
Bump! Any chance someone can review this or merge it! Thanks!

thedrow:
Sorry. I missed this PR.

PR #455, opened 2016-12-20
Fixes bug with 'prompt' parameter

bjmc:
This changeset fixes a bug with the OpenIDConnect `prompt=` parameter that occurs in the second invocation of the `openid_authorization_validator()` during a POST to the authorization endpoint. If `credentials['prompt']` is already a list, then calling `split()` triggers an `AttributeError`

This is a deliberately conservative fix that just avoids trying to `split()` twice. It's possible that this bug points to a deeper problem with this approach (are we sure we should be splitting `prompt` when returning the `request_info`?) in the OIDC parameter handling.  

Is this same bug likely to occur with the `ui_locales` parameter?


bjmc:
@thedrow Thoughts on this? This is an actual bug we've encountered. I could work around it on our end, but it seems like it might warrant fixing upstream.

thedrow:
Sorry this took so long. I'm releasing this as a new version.

PR #456, opened 2017-01-05
Add log statements to except clauses

empiricalthought:
These log statements would help me not tear my hair out while implementing LTI tool providers.

### DIFF
--- a/oauthlib/oauth1/rfc5849/endpoints/signature_only.py
+++ b/oauthlib/oauth1/rfc5849/endpoints/signature_only.py
@@ -34,13 +34,17 @@ class SignatureOnlyEndpoint(BaseEndpoint):
         """
         try:
             request = self._create_request(uri, http_method, body, headers)
-        except errors.OAuth1Error:
+        except errors.OAuth1Error as e:
+            log.info(
+                'Exception caught while validating request, %s.' % e)
             return False, None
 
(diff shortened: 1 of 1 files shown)


oauthlib/oauth1/rfc5849/endpoints/signature_only.py
```diff
@@ -34,13 +34,17 @@ def validate_request(self, uri, http_method='GET',
         """
         try:
             request = self._create_request(uri, http_method, body, headers)
-        except errors.OAuth1Error:
+        except errors.OAuth1Error as e:
```
bjmc: Nitpick: single-letter variable names are hard to grep for. Consider `err` or `error` instead?

empiricalthought: Sure -- I was following the convention(?) in https://github.com/idan/oauthlib/blob/67f973ff7f98bb3d892a33eda67ba1dab3bddead/oauthlib/oauth2/rfc6749/endpoints/base.py#L58 though -- down to the logging message.

bjmc: Yeah, I don't love that one, either. But let sleeping dogs lie.

Thanks for changing it.

PR #458, opened 2017-01-20
Documented PyJWT dependency and improved logging and exception messages

hoylen:
Updated documentation so it is clear that PyJWT (and not any jwt library) is needed. Corrected "pip install" command to install "pyjwt" instead of "jwt".

Added more information to the `NotImplemented` exception messages. Instead of saying "Subclasses must implement this function", it is now more useful because it says "Missing function implementation in className: nameOfMissingFunction".

Extra logging has been added to help in debugging. During development, the _signature base string_ signed by the client may be different from the one independently calculated by the server: logging these string helps debug such problems.

In _oauthlib/oauth1/rfc5849/endpoints/signature_only.py_ exceptions are caught and then False is simply returned. The details in those exceptions are useful for debugging what went wrong. Their values are now logged, rather than simply discarded.

### DIFF
--- a/docs/oauth1/server.rst
+++ b/docs/oauth1/server.rst
@@ -2,7 +2,7 @@
 Creating a Provider
 ===================
 
-OAuthLib is a dependency free library that may be used with any web
+OAuthLib is a framework independent library that may be used with any web
 framework. That said, there are framework specific helper libraries
 to make your life easier.
 
--- a/docs/oauth1/client.rst
+++ b/docs/oauth1/client.rst
@@ -64,15 +64,16 @@ Using the Client
 
 **RSA Signatures**
 
-    OAuthLib supports the 'RSA-SHA1' signature but does not install the jwt or
-    cryptography dependency by default. The cryptography package is much better
-    supported on Windows and Mac OS X than PyCrypto, and simpler to install.
-    OAuthLib uses the jwt package to smooth out its internal code.
-    Users can install cryptography using pip::
+    OAuthLib supports 'RSA-SHA1' signatures, but does not install the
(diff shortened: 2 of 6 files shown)


docs/oauth1/server.rst
```diff
@@ -2,7 +2,7 @@
 Creating a Provider
 ===================
 
-OAuthLib is a dependency free library that may be used with any web
+OAuthLib is a framework independent library that may be used with any web
```
thedrow: Why did you change this?

hoylen: Because OAuthLib has dependencies, so it is not "dependency free". The description of "dependency free" would mean it does not need any other package to run. OAuthLib (as indicated in the _requirements.txt_ file) needs PyJWT, blinker and cryptography packages to function.

I'm trying to guess what the original author really meant, and assume they meant to say it can be used with any framework (e.g. doesn't have to be Django) rather than "dependency free" (which it is clearly not, since it has dependencies).

thedrow:
Can you please rebase this branch?

thedrow:
Looks like the rebase went really wrong here. Can you try again?

hoylen:
Are you referring to the red cross next to 13387a4? There was a bug in my code, so it was failing the automatic checks. That was then fixed in b3f4cb9.

thedrow:
I'm referring to the entire code history in this PR that isn't related. See https://github.com/idan/oauthlib/pull/458#commits-pushed-bf2f9db

hoylen:
Done. This should be better.

thedrow:
Thanks!

PR #461, opened 2017-02-14
Add two error codes from RFC 6750

skion:
Add two error codes from [RFC 6750](https://tools.ietf.org/html/rfc6750) and align docstring formatting in that file.


thedrow:
Please rebase.

skion:
@thedrow All set.

PR #465, opened 2017-02-23
Hybrid response types should be fragment-encoded

skion:
According to the [multiple response types specification](http://openid.net/specs/oauth-v2-multiple-response-types-1_0-03.html#anchor5) all hybrid response types should return fragment-encoded responses.

However, according to the unit tests there might be some side effects.


thedrow:
If you want me to consider merging this, we need to understand the implications on existing clients/servers and fix the unit tests.

skion:
Yes, I'll have a look at what tests are affected by this and why.

skion:
@thedrow It created some problems with the inheritance of the AuthCode tests in the Hybrid flow tests. I've worked around that by making response modes more explicit in the tests.

thedrow:
Thanks!

PR #466, opened 2017-02-23
Improve prompt parameter validation

skion:
- Strip leading and trailing spaces.
- Disallow prompt=none with other values as per spec.
- Pass to credentials as a set.

I'm wondering if it would be better to also not enforce `id_token_hint` when `prompt=none` (the specification allows both), as for a developer it would be harder to eliminate this check in the default server than it is to add the constraint externally.


thedrow:
Can you please rebase? Do the existing tests cover your changes?

skion:
@thedrow Added test and rebased on top of #455.

thedrow:
Thanks!

PR #468, opened 2017-03-08
Changed all implicit imports * with explicit classes imports

TizianoPerrucci:
While integrating oauthlib with a Jython application we've found that Jython 2.7.0 has issues with `import *`.

This way of importing is in general strongly discouraged due to namespace pollution and unattended conflicts which could happen.

We've found only 3 occurrences of import * and we fixed all of them:
```
$ grep -R "from .* import \*" *
oauth1/__init__.py:from .rfc5849.errors import *
oauth2/__init__.py:from .rfc6749.errors import *
oauth2/rfc6749/clients/__init__.py:from .base import *
```

Please consider merging this PR and include in the next release.

Thank you.


Voronenko:
@idan Any chances to get it merged and released?  This is blocker currently, when you try to use library with Azure under Jython

thedrow:
I'll have to review this more closely.

wiliamsouza:
LGTM :+1:

PR #473, opened 2017-04-16
Replace all uses of assertEquals with assertEqual

jdufresne:
assertEquals is deprecated in favor of assertEqual.

When running tests, fixes warnings of the form:

  DeprecationWarning: Please use assertEqual instead.

For more information, see:

https://docs.python.org/3/library/unittest.html#deprecated-aliases


thedrow:
Thanks!

PR #474, opened 2017-04-16
Fix BytesWarning issued when using a string placeholder for bytes object

jdufresne:
When running tests, fixes warnings of the form:

.../oauthlib/oauthlib/oauth1/rfc5849/utils.py:56: BytesWarning: str() on a bytes instance


thedrow:
Thanks!

PR #475, opened 2017-04-26
Remove check on empty scopes for implicit grant flow

akira-dev:
I don't know why this test is made only in implicit grant type flow. Because in all others grant type flow if you don't provide scopes to get a token, token is generated with default_scopes of the client without errors. Moreover if you don't make this test in implicit grant type flow the behavior of this flow is the same as other grant type flows: it generate token with client default_scopes.


wiliamsouza:
Hi, if you take a look in the following test cases list you will see that this is not covered by any test:

https://github.com/idan/oauthlib/blob/master/tests/oauth2/rfc6749/grant_types/test_implicit.py#L28-L41
https://github.com/idan/oauthlib/blob/master/tests/oauth2/rfc6749/grant_types/test_authorization_code.py#L103-L136
https://github.com/idan/oauthlib/blob/master/tests/oauth2/rfc6749/grant_types/test_openid_connect.py#L141-L168
https://github.com/idan/oauthlib/blob/master/tests/oauth2/rfc6749/grant_types/test_resource_owner_password.py#L33-L90
https://github.com/idan/oauthlib/blob/master/tests/oauth2/rfc6749/grant_types/test_client_credentials.py#L50-L60

Add this tests is important to this PR can you provide that? Ensure the existence of a test that in the case of not providing a scope the default defined in `RequestValidator` is used.

duaneking:
@akira-dev  This build is failing.  If you are still interested in this contribution please update your base and merge so that tests pass.  If not. let us know and we will include it in our cleanup.  Thanks!

PR #480, opened 2017-06-14
redirect_uri mismatch

orenmazor:
One of our customers came across this issue. they provided the correct redirect_uri on the auth code request, but then the wrong one on the token request.

the error they receive was a 401 "Access Denied", which is correct, but not descriptive. I could start overriding things up in the stack to add a check, but I feel like a more descriptive error can just come back from here, seeing as how there's already an exception specifically for this.

thoughts? 

/cc @bjmc @thedrow

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
@@ -424,7 +424,7 @@ class AuthorizationCodeGrant(GrantTypeBase):
                                                            request.redirect_uri, request.client):
             log.debug('Redirect_uri (%r) invalid for client %r (%r).',
                       request.redirect_uri, request.client_id, request.client)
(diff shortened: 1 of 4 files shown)


orenmazor:
if this is acceptable i will happily adjust the test suite to handle it, obviously

bjmc:
This seems sensible to me, unless there's something in the OAuth2 spec that forbids giving more detailed information for security reasons. And this could be potentially a breaking change for clients that look for the old, generic error.

bjmc:
Actually, should this be `MismatchingRedirectURIError`? The context here is that the client sent a different redirect_uri during the authorization request than they sent on the token request.

orenmazor:
Yeah, that actually makes sense to me. updated.

thedrow:
Thanks for the contribution. It will be released soon.

PR #481, opened 2017-07-28
Pass through nonce in code flow

skion:
Nonces are *optional* in `authorization_code` flow, but still should be passed through if the RP provided one.

> If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token

Also the claims parameter appeared to be processed twice.

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
@@ -306,6 +306,7 @@ class OpenIDConnectBase(object):
 
         request_info = {
             'display': request.display,
+            'nonce': request.nonce,
             'prompt': prompt,
             'ui_locales': request.ui_locales.split() if request.ui_locales else [],
             'id_token_hint': request.id_token_hint,
@@ -336,9 +337,7 @@ class OpenIDConnectBase(object):
             desc = 'Request is missing mandatory nonce parameter.'
             raise InvalidRequestError(request=request, description=desc)
 
-        self._inflate_claims(request)
-
-        return {'nonce': request.nonce, 'claims': request.claims}
+        return {}
 
 
 class OpenIDConnectAuthCode(OpenIDConnectBase):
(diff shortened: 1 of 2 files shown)


thedrow:
How come the tests did pass before that? This implies we're missing coverage for a code branch where nonces are not present.
Can you please add a unit test for this code branch?

skion:
Reopening accidental close...

wiliamsouza:
`nonce` is no longer required as stated by http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

```
nonce
OPTIONAL. String value used to associate a Client session with an ID Token, and to mitigate replay attacks. The value is passed through unmodified from the Authentication Request to the ID Token. Sufficient entropy MUST be present in the nonce values used to prevent attackers from guessing values. For implementation notes, see Section 15.5.2.
```

skion:
@wiliamsouza It's always been optional in code flow, but should still be included if present.

wiliamsouza:
With missing tests LGTM

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -307,6 +307,7 @@ def openid_authorization_validator(self, request):
```
wiliamsouza: Suggestion: Add the following text to method doc string:

```
+        nonce
+                OPTIONAL. String value used to associate a Client session with an
+                ID Token, and to mitigate replay attacks. The value is passed
+                through unmodified from the Authentication Request to the ID Token.
+                Sufficient entropy MUST be present in the nonce values used to
+                prevent attackers from guessing values

```

skion:
>  Can you please add a unit test for this code branch?

@thedrow Something like this 13c6cf5?



Also, should the other OIDC params end up as credentials as well? If so, then can add them in similar way.

thedrow:
@skion Can you please rebase this branch?
Also, what other OIDC parameters are we talking about?

skion:
I think I did not spot `max_age` and `acr_values` anywhere, but we may also want things like `claims_locales`, `request` or `request_uri` eventually? 

Rebased!

thedrow:
Frankly I'm not familiar with OpenID as much as OAuth2 to tell you yes or no. Up to you.

PR #484, opened 2017-09-06
OpenID connect improvements 

wiliamsouza:
This pull request removes the need of using `grant_type=openid` in token endpoint and when defining a application credential the `authorization_grant_type` can be same already used (`authorization-code`) both for OpenID Connect and OAuth2.

OpenID Connect and OAuth2 specification links:

* http://openid.net/specs/openid-connect-core-1_0.html#TokenRequest
* http://openid.net/specs/openid-connect-core-1_0.html#HybridTokenEndpoint
* https://tools.ietf.org/html/rfc6749#section-4.1.3

To achieve this a new method `get_authorization_code_scopes` have to be added to `RequestValidator` along side news dispatches for implicit and token grant.

Example old way to send a token request:

```
curl -X POST \
    -H "Cache-Control: no-cache" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    "http://127.0.0.1:8000/o/token/" \
    -d "client_id=qLSq5KfSvYRho..." \
    -d "client_secret=bhOWtOxm1..." \
    -d "code=B8J3PQBpDeHFohgrVnvs26KrTAMNhI" \
    -d "redirect_uri=http://localhost/callback" \
    -d "grant_type=openid"
```

Example  fixed way to send a token request:

```
curl -X POST \
    -H "Cache-Control: no-cache" \
    -H "Content-Type: application/x-www-form-urlencoded" \
    "http://127.0.0.1:8000/o/token/" \
    -d "client_id=qLSq5KfSvYRho..." \
    -d "client_secret=bhOWtOxm1..." \
    -d "code=B8J3PQBpDeHFohgrVnvs26KrTAMNhI" \
    -d "redirect_uri=http://localhost/callback" \
    -d "grant_type=authorization_code"
``` 

I successfully tested using django-oauth-toolkit the following flows: 

* Implicit Flow
  * Returning only access_token
  * Returning only id_token
  *  Return access_token and id_token
* Authorization Flow
  * Returning only access_token
  * Returning id_token
* Hybrid flow
  * Returning code and id_token
  * Returning code and access_token
  * Returning code, id_token and access_token
* Client credentials
* Resource owner password based

Adding support to OpenID Connect is a working in progress based on this pull request.


skion:
Nice, I also ran into some missing bits in this area but worked around them in a different way. I'll need to dig up what I did exactly, but do believe we need something that looks like this PR.

thedrow:
@wiliamsouza Can you please rebase and fix the conflicts?

wiliamsouza:
@thedrow Done!

oauthlib/oauth2/rfc6749/grant_types/implicit.py
```diff
@@ -239,7 +238,12 @@ def create_token_response(self, request, token_handler):
 
         for modifier in self._token_modifiers:
             token = modifier(token, token_handler, request)
-        self.request_validator.save_token(token, request)
+
+        # In OIDC implicit flow it is possible to have a request_type that does
+        # not include the access_token! In this case there is no need to save a token.
+        if "token" in request.response_type.split():
```
thedrow: Kinda expensive to split the string into a list just to check if token is in it.

tests/oauth2/rfc6749/grant_types/test_openid_connect.py
```diff
@@ -289,3 +296,105 @@ def setUp(self):
...
+        self.request_validator.get_authorization_code_scopes.return_value = ('hello', 'openid')
+        self.dispatcher = AuthTokenGrantDispatcher(
+            self.request_validator,
+            default_token_grant=self.auth_grant,
+            oidc_token_grant=self.openid_connect_auth
+        )
+
+    def test_create_token_response_openid_without_code(self):
+        handler = self.dispatcher._handler_for_request(self.request)
+        self.assertTrue(isinstance(handler, AuthorizationCodeGrant))
```
skion: Is it possible to assert here (and below) how the mocked `get_authorization_code_scopes()` method was called?

oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
```diff
@@ -49,33 +52,37 @@ def __init__(self, request_validator, token_expires_in=None,
...
                                        response_types={
                                            'code': auth_grant_choice,
-                                           'token': implicit_grant,
-                                           'id_token': openid_connect_implicit,
-                                           'id_token token': openid_connect_implicit,
-                                           'code token': openid_connect_auth,
-                                           'code id_token': openid_connect_auth,
-                                           'code token id_token': openid_connect_auth,
+                                           'token': implicit_grant_choice,
+                                           'id_token': implicit_grant_choice,
```
skion: Should `id_token` and `id_token token` not point to the `openid_connect_implicit` grant directly?

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -76,6 +76,65 @@ def validate_authorization_request(self, request):
...
+    def __init__(self, request_validator, default_token_grant=None, oidc_token_grant=None):
+        self.default_token_grant = default_token_grant
+        self.oidc_token_grant = oidc_token_grant
+        self.request_validator = request_validator
+
+    def _handler_for_request(self, request):
+        handler = self.default_token_grant
+        scopes = ()
+        parameters = dict(request.decoded_body)
+        client_id = parameters.get('client_id', '')
```
skion: Why do `client_id` and `redirect_uri` default to an empty string instead of `None`?

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -76,6 +76,65 @@ def validate_authorization_request(self, request):
...
+        code = parameters.get('code', None)
+        redirect_uri = parameters.get('redirect_uri', '')
+
+        # If code is not pressent fallback to `default_token_grant` wich will
+        # rase an error for the missing `code` in `create_token_response` step.
```
skion: Typo: `rase`

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
wiliamsouza: Nothing special, changed to `None`, thanks.

wiliamsouza:
@thedrow @skion suggestions done. Thanks for review.

tests/oauth2/rfc6749/grant_types/test_openid_connect.py
wiliamsouza: Done in 34aa9b0

thedrow:
Is this ready for merge?

wiliamsouza:
@skion can answer the question!

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
```diff
@@ -76,6 +76,65 @@ def validate_authorization_request(self, request):
...
+        scopes = ()
+        parameters = dict(request.decoded_body)
+        client_id = parameters.get('client_id', None)
+        code = parameters.get('code', None)
+        redirect_uri = parameters.get('redirect_uri', None)
+
+        # If code is not pressent fallback to `default_token_grant` wich will
+        # raise an error for the missing `code` in `create_token_response` step.
+        if code:
+            scopes = self.request_validator.get_authorization_code_scopes(client_id, code, redirect_uri)
```
skion: What I don't particularly like, but also not really sure how to do differently, is that this forces to lookup the grant twice; once in `get_authorization_code_scopes()` and once in `validate_code()`. Not ideal for performance.

I wonder if we should pass in the `request` object here, as with many of the other validator methods. Then at least one can choose to attach the result of the first lookup to the request.

wiliamsouza: This is a gap between stateless endpoints:

* Authorization Endpoint
* Token Endpoint

`request_validator.get_authorization_code_scopes()` is called before `request_validator.validate_code()` which is used inside `AuthorizationCodeGrant.validate_token_request()` we can call `request_validator.validate_code()` here and pass in `request.code_validate` a `True` value indicating the code was already validated and in `AuthorizationCodeGrant.validate_token_request()` line 410 add a check before call `request_validator.validate_code()` again.

oauthlib/oauth2/rfc6749/grant_types/implicit.py
wiliamsouza: `split()` can not be removed because:

```
ipdb> 'token' in 'id_token'
True
ipdb> 'token' in ['id_token']
False
```
When using implicit flow:

```
http://127.0.0.1:8000/o/authorize?response_type=id_token&client_id=Ktn0Sh4hO2gA8PKC2aqsauY4ZCxNyIdF1wNfFfJ3&redirect_uri=http://localhost/callback&scope=openid&state=af0ifjsldkj&nonce=n-0S6_WzA2Mj
```
without `split` it return access_token instead of only returning `id_token` as requested.

oauthlib/oauth2/rfc6749/grant_types/openid_connect.py
skion: I guess what I meant was: Could we use the following signature instead?

    def get_authorization_code_scopes(self, client_id, code, redirect_uri, request):

This seems to be a common pattern already, and it would allow us to cache database lookups on the `request` object easily across validation calls.

wiliamsouza: @skion done

skion:
LGTM!

wiliamsouza:
@thedrow ready for merge!

thedrow:
Thanks. Is this a breaking change? Just so I'll know how to tag the release.

wiliamsouza:
@thedrow Yes not backward compatible.

skion:
Turns out this [leads to problems](https://github.com/idan/oauthlib/pull/463#issuecomment-337722389) with existing code that doesn't have `request_validator.get_authorization_code_scopes` defined, as this PR expects it to exist and return something sensible.

To fix we could catch the `NotImplementedError` in the dispatcher, or alternatively just return an empty list from the base request validator. But neither is particularly aesthetically pleasing.

thedrow:
Oh this should have been released as 3.x :(

lepture:
@thedrow we need to add a `breaking changes` tag.

duaneking:
Can we please back this out? It's doing more harm than good right now.

I love the idea of OpenIDConnect support, but I also think it should be optional and not shoved into the library without consideration of servers/providers/clients that don't want it or don't need it.

My opinion is to let the OAuth2 library do Oauth2, add the extras openid connect wants on top with a library of its own that uses the openid2 library, and keep a good separation of concerns so the code stays as SOLID and DRY as possible.

skion:
@duaneking Not so easy to back out OIDC entirely, since its support was added long before this PR already, and mostly modular/separate from the existing code.

The exception being the preconfigured server code, which I agree would have been better to leave alone and add a dedicated OIDC server. But undoing this now would also be a breaking change.

IMO this PR should not have been a breaking change at all, and my suggestion would be: Find a fix to make this non-breaking and release 2.0.6, and plan for more structural changes in 3.0.

wiliamsouza:
There is no way to remove 2.0.5 from the history?

lepture:
@wiliamsouza no. But there is a 2.0.6 in pypi now. I cherry picked every commits except this one.

wiliamsouza:
@lepture Great job!

thedrow:
Can we reopen this PR or release 3.0 with the relevant commit?

wiliamsouza:
@thedrow After https://github.com/idan/oauthlib/pull/488 we can release 3.0

skion:
@wiliamsouza Would this be backward compatible if we just add [this line 78](https://github.com/oauthlib/oauthlib/pull/484/files#diff-450ebe58a6668300682243653b4df952L78) back in?

PR #487, opened 2017-09-22
initialization of variable

MitterdorferMathias:
if `in_uri` is called and `response_mode` is not initialized an exception is raised (happen if used with e.g. Flask_OAuthlib)


thedrow:
Thanks. Feel free to add yourself to the AUTHORS file.

skion:
Thanks @MitterdorferMathias :+1:

Issue #50, opened 2012-07-21
OAuth2 Server - JWT token support


ib-lundgren:
As per http://tools.ietf.org/html/draft-jones-oauth-jwt-bearer-04.

bigblind:
Is this still being worked on?

thedrow:
@bigblind It was recently implemented.
However this issue should stay open until the JWT token RFC is finalized.

LucasRoesler:
@thedrow If this has been implement, is there any documentation on how to use it?  The RTD simply points here and nothing in the source code seems to jump out to me.

thedrow:
There is no documentation currently since the standard is not yet finalized.

joar:
Unless I'm mistaken it seems that the standard is finalized: https://tools.ietf.org/html/rfc7523

thedrow:
It is now a proposed standard which means it is being finalized. Unless something unexpected happens it will be finalized in the coming months. I think it's safe to say that we can start implementing it. Any volunteers?

Antherkiv:
Hi @thedrow, the jwt it's implemented, it's ok?, do you know if exists a django provider that use it?, thanks.

thedrow:
@Antherkiv The current implementation is according to the 4th draft of the specification. Someone needs to bring it up to speed with the current finalized specification.
I do not know any Django providers that use it.

clintonb:
There seems to be some confusion. I'm reading here that folks believe a *server* has been implemented. I only see a client: [ServiceApplicationClient](https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/clients/service_application.py). If a server has indeed been implemented, would someone point me to it?

thedrow:
@clintonb Seems like you're right. The server is not implemented.

KptMarchewa:
Any work done on this?

thedrow:
I have no need for this feature for now so no.
If you need it, feel free to issue a PR.

atif1996:
We could definitely use this feature.  Would this be useful for anyone else?

philloooo:
[...]

PR #488, opened 2017-10-03
Openid connect jwt

wiliamsouza:
This pull request adds support to receive JWT token in request in the following form:

```
curl -vv --header "Content-Type: application/json" \
         --header "Accept: application/json; indent=4" \
         --header "Authorization: Bearer eyJhbGciOiAiUlMyNTYifQ..." \
         http://127.0.0.1:8000/v1/users/
...
> GET /v1/users/ HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.53.1
> Content-Type: application/json
> Accept: application/json; indent=4
> Authorization: Bearer eyJhbGciOiAiUlMyNTYifQ....
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Date: Sun, 01 Oct 2017 19:53:52 GMT
< Server: WSGIServer/0.2 CPython/3.6.2
< Content-Type: application/json
< Vary: Accept
< Allow: GET, POST, HEAD, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 159
< 
[
    {
        "url": "http://127.0.0.1:8000/v1/users/1/",
        "username": "wiliam",
        "email": "wiliam@olist.com",
        "is_staff": true
    }
* Closing connection 0
]
```
Don't find tests for `tokens.py` model. Any clue?

### DIFF
--- a/oauthlib/oauth2/rfc6749/request_validator.py
+++ b/oauthlib/oauth2/rfc6749/request_validator.py
@@ -312,8 +312,24 @@ class RequestValidator(object):
         """
         raise NotImplementedError('Subclasses must implement this method.')
 
-    def get_id_token(self, token, token_handler, request):
+    def get_jwt_bearer_token(self, token, token_handler, request):
+        """Get JWT Bearer token or OpenID Connect ID token
+
+        If using OpenID Connect this SHOULD call `oauthlib.oauth2.RequestValidator.get_id_token`
+
+        :param token: A Bearer token dict
+        :param token_handler: the token handler (BearerToken class)
+        :param request: the HTTP Request (oauthlib.common.Request)
+        :return: The JWT Bearer token or OpenID Connect ID token (a JWS signed JWT)
+
+        Method is used by JWT Bearer and OpenID Connect tokens:
+            - JWTToken.create_token
         """
+        raise NotImplementedError('Subclasses must implement this method.')
+
+    def get_id_token(self, token, token_handler, request):
+        """Get OpenID Connect ID token
+
         In the OpenID Connect workflows when an ID Token is requested this method is called.
         Subclasses should implement the construction, signing and optional encryption of the
         ID Token as described in the OpenID Connect spec.
@@ -344,12 +360,31 @@ class RequestValidator(object):
         # the request.scope should be used by the get_id_token() method to determine which claims to include in the resulting id_token
         raise NotImplementedError('Subclasses must implement this method.')
 
+    def validate_jwt_bearer_token(self, token, scopes, request):
+        """Ensure the JWT Bearer token or OpenID Connect ID token are valids and authorized access to scopes.
+
+        If using OpenID Connect this SHOULD call `oauthlib.oauth2.RequestValidator.get_id_token`
+
+        OpenID connect core 1.0 describe how to validate an id_token:
+            - http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#HybridIDTValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#HybridIDTValidation2
+
+        :param token: Unicode Bearer token
+        :param scopes: List of scopes (defined by you)
+        :param request: The HTTP Request (oauthlib.common.Request)
+        :rtype: True or False
+
+        Method is indirectly used by all core OpenID connect JWT token issuing grant types:
+            - Authorization Code Grant
+            - Implicit Grant
+            - Hybrid Grant
+        """
+        raise NotImplementedError('Subclasses must implement this method.')
+
     def validate_id_token(self, token, scopes, request):
         """Ensure the id token is valid and authorized access to scopes.
-
-        :param token: A string of JWT serialized.
-        :param scopes: A list of scopes associated with the protected resource.
-        :param request: The HTTP Request (oauthlib.common.Request)
 
         OpenID connect core 1.0 describe how to validate an id_token:
             - http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
--- a/oauthlib/oauth2/rfc6749/request_validator.py
+++ b/oauthlib/oauth2/rfc6749/request_validator.py
@@ -365,6 +365,8 @@ class RequestValidator(object):
 
         If using OpenID Connect this SHOULD call `oauthlib.oauth2.RequestValidator.get_id_token`
 
+        If not using OpenID Connect this can `return None` to avoid 5xx rather 401/3 response.
+
         OpenID connect core 1.0 describe how to validate an id_token:
             - http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
             - http://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation
--- a/oauthlib/oauth2/rfc6749/request_validator.py
+++ b/oauthlib/oauth2/rfc6749/request_validator.py
@@ -344,6 +344,31 @@ class RequestValidator(object):
         # the request.scope should be used by the get_id_token() method to determine which claims to include in the resulting id_token
         raise NotImplementedError('Subclasses must implement this method.')
 
+    def validate_id_token(self, token, scopes, request):
+        """Ensure the id token is valid and authorized access to scopes.
+
+        :param token: A string of JWT serialized.
+        :param scopes: A list of scopes associated with the protected resource.
+        :param request: The HTTP Request (oauthlib.common.Request)
+
+        OpenID connect core 1.0 describe how to validate an id_token:
+            - http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#ImplicitIDTValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#HybridIDTValidation
+            - http://openid.net/specs/openid-connect-core-1_0.html#HybridIDTValidation2
+
+        :param token: Unicode Bearer token
+        :param scopes: List of scopes (defined by you)
+        :param request: The HTTP Request (oauthlib.common.Request)
+        :rtype: True or False
+
+        Method is indirectly used by all core OpenID connect JWT token issuing grant types:
+            - Authorization Code Grant
+            - Implicit Grant
+            - Hybrid Grant
+        """
+        raise NotImplementedError('Subclasses must implement this method.')
+
     def validate_bearer_token(self, token, scopes, request):
         """Ensure the Bearer token is valid and authorized access to scopes.
 
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -328,7 +328,7 @@ class JWTToken(TokenBase):
 
         request.expires_in = expires_in
 
-        return self.request_validator.get_id_token(None, None, request)
+        return self.request_validator.get_jwt_bearer_token(None, None, request)
 
     def validate_request(self, request):
         token = None
@@ -336,7 +336,7 @@ class JWTToken(TokenBase):
             token = request.headers.get('Authorization')[7:]
         else:
             token = request.access_token
-        return self.request_validator.validate_id_token(
+        return self.request_validator.validate_jwt_bearer_token(
             token, request.scopes, request)
 
     def estimate_type(self, request):
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -341,7 +341,7 @@ class JWTToken(TokenBase):
 
     def estimate_type(self, request):
         token = request.headers.get('Authorization', '')[7:]
-        if token.count('.') in (2, 4):
+        if token.startswith('ey') and token.count('.') in (2, 4):
             return 10
         else:
             return 0
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -24,8 +24,6 @@ except ImportError:
     from urllib.parse import urlparse
 
 
-
-
 class OAuth2Token(dict):
 
     def __init__(self, params, old_scope=None):
@@ -301,5 +299,49 @@ class BearerToken(TokenBase):
             return 9
         elif request.access_token is not None:
             return 5
+        else:
+            return 0
+
+
+class JWTToken(TokenBase):
+    __slots__ = (
+        'request_validator', 'token_generator',
+        'refresh_token_generator', 'expires_in'
+    )
+
+    def __init__(self, request_validator=None, token_generator=None,
+                 expires_in=None, refresh_token_generator=None):
+        self.request_validator = request_validator
+        self.token_generator = token_generator or random_token_generator
+        self.refresh_token_generator = (
+            refresh_token_generator or self.token_generator
+        )
+        self.expires_in = expires_in or 3600
+
+    def create_token(self, request, refresh_token=False, save_token=False):
+        """Create a JWT Token, using requestvalidator method."""
+
+        if callable(self.expires_in):
+            expires_in = self.expires_in(request)
+        else:
+            expires_in = self.expires_in
+
+        request.expires_in = expires_in
+
+        return self.request_validator.get_id_token(None, None, request)
+
+    def validate_request(self, request):
+        token = None
+        if 'Authorization' in request.headers:
+            token = request.headers.get('Authorization')[7:]
+        else:
+            token = request.access_token
+        return self.request_validator.validate_id_token(
+            token, request.scopes, request)
+
+    def estimate_type(self, request):
+        token = request.headers.get('Authorization', '')[7:]
+        if token.count('.') in (2, 4):
+            return 10
         else:
             return 0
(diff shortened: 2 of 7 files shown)


wiliamsouza:
@thedrow it would be great to have these changes in the next release.

thedrow:
If there are no existing tests could you cover only what you added? If you have the time to cover all the module that would be great.

skion:
I wonder if it is better to unlink ID Tokens from (more general) JWT tokens here...

Some providers generate ID Tokens fit for one-time validation with an expiry that cover just the login process. And such providers may want to issue JWT-structured access tokens instead. 

I'm guessing the validation semantics would be largely the same, but the naming might be confusing in this situation?

wiliamsouza:
The suggestion is rename `JWTToken` to `IDToken`?

skion:
No, the other way around...

duaneking:
What is the status of this?

duaneking:
Can we please get a update on the progress of this?

I can see this as helping the transition to supporting OIDC and it looks like a lot of work for adding non-openidconenct JWT support per #50 may be shared by this.

thedrow:
@wiliamsouza ping?

duaneking:
To be clear: I think this will help to support RFC7519

wiliamsouza:
@thedrow Missing time to write the tests (If someone want help feel free).

I will get back to openid connect next month as a professional job priority.

@duaneking This is part of a effort to add oidc support to django-oauth-toolkit [1] I'm not sure where the code necessary to create JWT should live so for now it's all  django-oauth-toolkit  side.

[1]  https://github.com/evonove/django-oauth-toolkit/compare/master...wiliamsouza:openid-connect

wiliamsouza:
@thedrow tests done.
@skion I Change my mind ID Token is a JWT with a fields defined in oidc spec and when referring to OpenID connect this seems to be a better name.

Kudos to @Peter-Slump for the tests!

thedrow:
I see that you added a method that raises a NotImplementedError. Is this PR a breaking change?

oauthlib/oauth2/rfc6749/request_validator.py
```diff
@@ -344,6 +344,31 @@ def get_id_token(self, token, token_handler, request):
         # the request.scope should be used by the get_id_token() method to determine which claims to include in the resulting id_token
         raise NotImplementedError('Subclasses must implement this method.')
 
+    def validate_id_token(self, token, scopes, request):
```
skion: I still wouldn't tie the naming here to ID Tokens specifically, as I wouldn't expect *all* JWT-based access tokens to be ID Tokens. Some servers might hand out special purpose JWT access tokens, next to their ID Tokens.

In that light, could this be named `validate_jwt_bearer_token` instead and have a companion `get_jwt_bearer_token` to create it?

Then people *could* alias those back to `get_id_token` if they want, but also have the option to split them up.

oauthlib/oauth2/rfc6749/tokens.py
```diff
@@ -303,3 +301,47 @@ def estimate_type(self, request):
...
+        """Create a JWT Token, using requestvalidator method."""
+
+        if callable(self.expires_in):
+            expires_in = self.expires_in(request)
+        else:
+            expires_in = self.expires_in
+
+        request.expires_in = expires_in
+
+        return self.request_validator.get_id_token(None, None, request)
```
skion: Similar to my other comment: What if someone wants to hand out JWT access tokens that are different from the ID Tokens they generate?

oauthlib/oauth2/rfc6749/tokens.py
```diff
@@ -303,3 +301,47 @@ def estimate_type(self, request):
...
+
+    def validate_request(self, request):
+        token = None
+        if 'Authorization' in request.headers:
+            token = request.headers.get('Authorization')[7:]
```
skion: Might be nice to align this with #491 already?

oauthlib/oauth2/rfc6749/tokens.py
```diff
@@ -303,3 +301,47 @@ def estimate_type(self, request):
...
+        else:
+            token = request.access_token
+        return self.request_validator.validate_id_token(
+            token, request.scopes, request)
+
+    def estimate_type(self, request):
+        token = request.headers.get('Authorization', '')[7:]
+        if token.count('.') in (2, 4):
```
skion: Curious: When would there be 4 segment separators?

Also, there's no rule saying a normal access token can't contain dots of course. Wondering if you should include a check for `ey` or `.ey` here perhaps.

wiliamsouza:
@thedrow This method is only required for OpenID connect this mean when you use `response_type=id_token` implementations that use only OAuth will not broke.

oauthlib/oauth2/rfc6749/request_validator.py
wiliamsouza: This effort is to add OpenID connect support and for this use case the name is correct. Add JWT token support to OAuth is a different thing and can be handled in near future and at this point all methods that handle ID token and JWT token can converge but think it's a future refactoring not part of this PR.

oauthlib/oauth2/rfc6749/tokens.py
wiliamsouza: The last to be merged resolve the conflict?

oauthlib/oauth2/rfc6749/tokens.py
wiliamsouza: I really don't remember... Will try to generate every accepted ID tokens by the specification to see a way to improve this tests

oauthlib/oauth2/rfc6749/request_validator.py
skion: Does OIDC specify somewhere how to use ID Tokens as JWT Bearer tokens?

To me it seems that you *have* in fact implemented generic (OAuth, not OIDC) JWT Bearer token support (as also the generic class name `JWTToken` would suggest), except for the token validator functions that have ID Token specific names.

Generalising this later would likely lead to an inconvenient upgrade path, since then `JWTToken` is already expected to call the OIDC-specific validators.

Unless I'm missing the point of the PR, I see no reason to keep this OIDC-specific detail in. Or is there?

wiliamsouza: @skion The OIDC don't say only:

* http://openid.net/specs/openid-connect-core-1_0.html#TokenResponse
* http://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse
* https://tools.ietf.org/html/rfc6750

Make sense rename `validate_id_token` to `validate_jwt_bearer_token`

oauthlib/oauth2/rfc6749/tokens.py
wiliamsouza: Will rename this call to `get_jwt_bearer_token` and let explicit that for OIDC this sould call `get_id_token`

oauthlib/oauth2/rfc6749/tokens.py
wiliamsouza: @skion Found it!

* https://tools.ietf.org/html/rfc7516#section-9

and for access token no period is used:

 https://github.com/idan/oauthlib/blob/7a0783d356504867f4a4868717f6e5e7d1b90a10/oauthlib/common.py#L194-L203
https://github.com/idan/oauthlib/blob/7a0783d356504867f4a4868717f6e5e7d1b90a10/oauthlib/common.py#L29-L31

skion: Ah, JWE's! :+1: 

About the non-JWT character set: Yes, that's how oauthlib generates tokens, but another server might still include periods in them. And someone might use oauthlib as a *client* against a 3rd party server...

That said, I'm not sure what the best way is either, just that checking for `.`'s feels a bit shallow.

wiliamsouza: When used as client oauthlib validate tokens before send it to 3rd party server?

It's guaranteed `ey` is added to every JWT token?

Can we let this check and see if it is a problem?

wiliamsouza:
@thedrow @skion PR updated

oauthlib/oauth2/rfc6749/tokens.py
skion: > When used as client oauthlib validate tokens before send it to 3rd party server?

Exactly

> It's guaranteed ey is added to every JWT token?

I tried this yesterday and could not thing of a reason why a JWT would **not** start with `{"`, but haven't thoroughly checked it either.

> Can we let this check and see if it is a problem?

Just beware that you'll go wrong with the first [example token](https://tools.ietf.org/html/rfc6750#section-2.1) in the RFC :fearful:

wiliamsouza: @skion I'm can't see other approach to solve this than validate the token at it early stage. What you think?

Some thing like:
```
>>> from jwcrypto import jwt
>>>jwt.JWT(jwt='token')
...
ValueError: Token format unrecognized

>>> jwt.JWT(jwt='t.oke.n')
...
InvalidJWSObject: Invalid JWS Object [Invalid format] {ValueError('Invalid base64 string',)}

>>> jwt.JWT(jwt='t.o.k.e.n')
...
InvalidJWEData: Invalid format {ValueError('Invalid base64 string',)}

>>> jwt.JWT(jwt='eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiYWRtaW4iOnRydWV9.TJVA95OrM7E2cBab30RMHrHDcEfxjoYZgeFONFh7HgQ')
<jwcrypto.jwt.JWT at 0x7f059df786d8>
```

skion: Yeah, that crossed my mind as well, but the tricky bit with it would be that you'd need to exclude all exceptions that might be raised for valid but expired or otherwise bad tokens.

And I also liked your PR in the way that the JWT logic itself was contained to the request validator, leaving the developer free to use whatever JWT library they might prefer.

Perhaps the following is a reasonable trade off that doesn't pull in a JWT lib as a dependency?

```python
if token.startswith('ey') and token.count('.') in (2, 4):
```

wiliamsouza: @skion Done! Thanks.

wiliamsouza:
@skion Need your ok here.

duaneking:
Thanks for the work on this @wiliamsouza !

In your mind how hard would it be to fully support RFC7519 (Per https://tools.ietf.org/html/rfc7519) after this change? I noticed your curl didnt have it, but thats a hard requirment for OIDC and that seems to be your goal here.

skion:
@wiliamsouza About @thedrow's question, I would think that existing code that happens to use `pre_configured.Server()` might indeed hit the `NotImplemented` in case a user sends a token that looks like a JWT? Or is this not the case.

Also, is by intention to add the JWT support only to the `pre_configured.Server()`, and not the other servers that support a resource endpoint?

wiliamsouza:
@duaneking What is the way to instruct to OAuth to use JWT? Add settings like `OAUTHLIB_USE_JWT_TOKEN=True`? If there is a issue for JWT support we can move that discussion to that isssue.

wiliamsouza:
@skion It will hit `NotImplemented` if token looks like JWT:
 https://github.com/idan/oauthlib/pull/488/files#diff-f1e3c9d49e6a9992e075c962b46b6f38R344
The chance a token start with ey and have two to four point is very difficult.

My suggestion is to create new issues to add support in `pre_configured.WebApplicationServer` and `pre_configured.MobileApplicationServer` in a separated pull request.

skion:
>  The chance a token start with ey and have two to four point is very difficult.

Someone could **intentionally** send a JWT though, in which case the `NotImplemented` would always be hit and (likely) cause a 5xx rather a 401/3 response. Not ideal I guess, even if we mark this as a breaking change.

Would there be a way to make the PR backward compatible? Perhaps if `get_jwt_bearer_token` in the request validator returns `None` by default instead of raising an exception?

wiliamsouza:
@skion seems to work but the used method is `validate_jwt_bearer_token` it respond ` 401 Unauthorized`:
```
curl -vv --header "Content-Type: application/json" --header "Accept: applicaton/json; indent=4" --header "Authorization: Bearer eyJhbGciOiAiUlMyNTYifQ.eyJpc3MiOiAiaHR0cHM6Ly9pZC5vbGlzdC5jb20iLCAic3ViIjogIjEiLCAiYXVkIjogIkt0bjBTaDRoTzJnQThQS0MyYXFzYXVZNFpDeE55SWRGMXdOZkZmSjMiLCAiZXhwIjogMTUxNTY0MjgzNSwgImlhdCI6IDE1MTU2MDY4MzUsICJhdXRoX3RpbWUiOiAxNTE1NDk3MzM3LCAibm9uY2UiOiAibi0wUzZfV3pBMk1qIiwgImF0X2hhc2giOiAiT0ROaE9HVmhNVGRtTW1RNU5qWTFaQT09In0.u0_tMWAxxKX3Ax9xy0ZwRW08vDwd2AIaIwFEo913d90x5DW59kStYTUu-Y2ri2Fq6qtrPovWugRC2Dyy7VZVgEdCbpSlPkbsg41lTdO2jG9HvoGI7j7eGEYwOtnrIdbVxn583Zdeol9IiUDvtCVG52MhyzfYFzA2CQ-XC6s-AEzzp9fwtYoToHNSXKzpqGTJwmLj3kf9bX9H4jgZ3sTF4m6Kh0NLiQo1GreKuavjfUwvGxXi5XHWNv1lSoFpdfFlN9LMUOnGvYzBwniusTf7PQh-S06j8UntRsfUyLDQtJY7c1an2-JP3YwnD0_K3UCDEzq9DCr-y37qWtLTwghvqw" --request GET http://127.0.0.1:8000/v1/users/
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> GET /v1/users/ HTTP/1.1
> Host: 127.0.0.1:8000
> User-Agent: curl/7.53.1
> Content-Type: application/json
> Accept: application/json; indent=4
> Authorization: Bearer eyJhbGciOiAiUlMyNTYifQ.eyJpc3MiOiAiaHR0cHM6Ly9pZC5vbGlzdC5jb20iLCAic3ViIjogIjEiLCAiYXVkIjogIkt0bjBTaDRoTzJnQThQS0MyYXFzYXVZNFpDeE55SWRGMXdOZkZmSjMiLCAiZXhwIjogMTUxNTY0MjgzNSwgImlhdCI6IDE1MTU2MDY4MzUsICJhdXRoX3RpbWUiOiAxNTE1NDk3MzM3LCAibm9uY2UiOiAibi0wUzZfV3pBMk1qIiwgImF0X2hhc2giOiAiT0ROaE9HVmhNVGRtTW1RNU5qWTFaQT09In0.u0_tMWAxxKX3Ax9xy0ZwRW08vDwd2AIaIwFEo913d90x5DW59kStYTUu-Y2ri2Fq6qtrPovWugRC2Dyy7VZVgEdCbpSlPkbsg41lTdO2jG9HvoGI7j7eGEYwOtnrIdbVxn583Zdeol9IiUDvtCVG52MhyzfYFzA2CQ-XC6s-AEzzp9fwtYoToHNSXKzpqGTJwmLj3kf9bX9H4jgZ3sTF4m6Kh0NLiQo1GreKuavjfUwvGxXi5XHWNv1lSoFpdfFlN9LMUOnGvYzBwniusTf7PQh-S06j8UntRsfUyLDQtJY7c1an2-JP3YwnD0_K3UCDEzq9DCr-y37qWtLTwghvqw
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 401 Unauthorized
< Date: Wed, 10 Jan 2018 17:55:08 GMT
< Server: WSGIServer/0.2 CPython/3.6.2
< Content-Type: application/json
< WWW-Authenticate: Bearer realm="api"
< Vary: Accept
< Allow: GET, POST, HEAD, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 65
< 
{
    "detail": "Authentication credentials were not provided."
* Closing connection 0

```
What is the best use default to `return None` or `NotImplemented` and document to use `return None` in `validate_jwt_bearer_token` if not using OpenID Connect support?

duaneking:
@wiliamsouza Right now the lib just seems to accept random data as the token.  Looking at the code its clear to me there was some assumptions made that will need to be refactored to support other token formats.  Your work is helping with that, but because you are doing it, it also means we/you have an opportunity to set the architectural tone for support.

My thinking is that the entire idea of a token format or style is something most to the system simply shouldn't care about; it should just be factored out generically so the system can be as agnostic about the token formats it consumes/validates as possible while still actually doing validation and enforcing a default for generation.  In the best case, it should support multiple formats at the same time for consumption/validation, depending on whats enabled, so you can do a seamless update/upgrade of token formats and not lose currently logged in users on a redeploy that enables a new token field or something.

In my mind JWT via #50 adds the most benefit over the current "simple random string token" format created out of random strings, so simply any effort to enable it is a positive.

wiliamsouza:
@thedrow Need your ok too.
@skion ping.

wiliamsouza:
@thedrow, @skion Ping! Don't let this PR die, please! With the OAuthLib new release we have the change o approval of this PR https://github.com/evonove/django-oauth-toolkit/pull/545 that adds support to django-oauth-toolkit

thedrow:
As far as I'm concerned this PR LGTM.
But I think you should have a look at @duaneking's comments and see if there's something you'd rather change.

skion:
Same for me!

wiliamsouza:
@skion Thanks!
@thedrow Can you merge this! Please :) What we need to release 3.0.0 as a backward incompatible?
@duaneking Can you provide a PR after this get merged about your refactor ideas?

thedrow:
@wiliamsouza Just the release notes. I'd appreciate if you'd release 3.0.0 yourself.

duaneking:
@wiliamsouza I just really don't want args to be switched around; if we have backward compatibility needs then args should stay in the same positional index and new ones should have sane but config overridden enabled defaults.

oauthlib/oauth2/rfc6749/tokens.py
MattBlack85: rebased and solved the conflict

PR #491, opened 2017-10-20
Check that the Bearer header is properly formatted

MattBlack85:
While testing some things we noticed that passing a header with a typo like `Beaver` makes the app behaving like it was `Bearer`. I don't know if this is a desired behaviour or not but it seems odd to me. I made the check a little bit more rigid and added test cases for that.
Feel free to close and kill this if this is an intended behaviour

### DIFF
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -220,6 +220,24 @@ def signed_token_generator(private_pem, **kwargs):
     return signed_token_generator
 
 
+def get_token_from_header(request):
+    """
+    Helper function to extract a token from the request header.
+    :param request: The request object
+    :return: Return the token or None if the Authorization header is malformed.
+    """
+    token = None
+
+    if 'Authorization' in request.headers:
+        split_header = request.headers.get('Authorization').split()
+        if len(split_header) == 2 and split_header[0] == 'Bearer':
+            token = split_header[1]
+    else:
+        token = request.access_token
+
+    return token
+
+
 class TokenBase(object):
 
     def __call__(self, request, refresh_token=False):
@@ -286,16 +304,12 @@ class BearerToken(TokenBase):
         return token
 
     def validate_request(self, request):
-        token = None
-        if 'Authorization' in request.headers:
-            token = request.headers.get('Authorization')[7:]
-        else:
-            token = request.access_token
+        token = get_token_from_header(request)
         return self.request_validator.validate_bearer_token(
             token, request.scopes, request)
 
     def estimate_type(self, request):
-        if request.headers.get('Authorization', '').startswith('Bearer'):
+        if request.headers.get('Authorization', '').split(' ')[0] == 'Bearer':
             return 9
         elif request.access_token is not None:
             return 5
@@ -331,17 +345,13 @@ class JWTToken(TokenBase):
         return self.request_validator.get_jwt_bearer_token(None, None, request)
 
     def validate_request(self, request):
-        token = None
-        if 'Authorization' in request.headers:
-            token = request.headers.get('Authorization')[7:]
-        else:
-            token = request.access_token
+        token = get_token_from_header(request)
         return self.request_validator.validate_jwt_bearer_token(
             token, request.scopes, request)
 
     def estimate_type(self, request):
-        token = request.headers.get('Authorization', '')[7:]
-        if token.startswith('ey') and token.count('.') in (2, 4):
+        split_header = request.headers.get('Authorization', '').split()
+
+        if len(split_header) == 2 and split_header[0] == 'Bearer' and split_header[1].startswith('ey') and split_header[1].count('.') in (2, 4):
             return 10
-        else:
-            return 0
+        return 0
--- a/tests/oauth2/rfc6749/test_tokens.py
+++ b/tests/oauth2/rfc6749/test_tokens.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 import mock
 
+from oauthlib.common import Request
 from oauthlib.oauth2.rfc6749.tokens import *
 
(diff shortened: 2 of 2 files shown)


tests/oauth2/rfc6749/test_tokens.py
```diff
@@ -59,9 +61,17 @@ class TokenTest(TestCase):
     bearer_headers = {
         'Authorization': 'Bearer vF9dft4qmT'
     }
+    fake_bearer_headers = {
+        'Authorization': 'Beaver vF9dft4qmT'
```
skion: Maybe add some further odd tokens? Such as

    BearerNoSpace
    Bearer Multi Space
    BeaverNoSpace
    Beaver Multi Space

MattBlack85: Can do

MattBlack85:
@skion I finally added those tests, it seems tho that travis is disabled for the repo now?

skion:
That's potentially due to the moving of the repo, stay tuned on #511.

I think you'll need to apply the same patch to [these](https://github.com/MattBlack85/oauthlib/blob/59f72697a4648e79c31a4850988d9ca96b5e2f5b/oauthlib/oauth2/rfc6749/tokens.py#L338) two [lines](https://github.com/MattBlack85/oauthlib/blob/59f72697a4648e79c31a4850988d9ca96b5e2f5b/oauthlib/oauth2/rfc6749/tokens.py#L345) still, not that #488 has been merged.

MattBlack85:
@skion thanks, I'll implement the check also for JWT

MattBlack85:
@skion I think the case `Bearer Multi Space` or `Bearer  DoubleSpace` are valid header and shouldn't fail the validation cause the header type is valid. Instead what will happen is that the token will fail the validation cause it will be `''` or a truncated token. What do you think of these cases?

MattBlack85:
sorry for the noise, the `.split()` will give more than 2 elements and it won't pass the validation anyway, my bad

skion:
FWIW: The Bearer token header is described like [this](https://tools.ietf.org/html/rfc6750#section-2.1):

    b64token    = 1*( ALPHA / DIGIT /
                      "-" / "." / "_" / "~" / "+" / "/" ) *"="
    credentials = "Bearer" 1*SP b64token

So multiple spaces after the `Bearer` keyword should not be a problem.

MattBlack85:
@skion updated my PR, everything should be ok now, let me know in case

oauthlib/oauth2/rfc6749/tokens.py
```diff
@@ -331,17 +345,13 @@ def create_token(self, request, refresh_token=False, save_token=False):
...
-        else:
-            token = request.access_token
+        token = get_token_from_header(request)
         return self.request_validator.validate_jwt_bearer_token(
             token, request.scopes, request)
 
     def estimate_type(self, request):
-        token = request.headers.get('Authorization', '')[7:]
-        if token.startswith('ey') and token.count('.') in (2, 4):
+        token_type, token, *_ = request.headers.get('Authorization', '').split(' ')
```
skion: That syntax won't fly on Python 2.7...

And should we use `.split()` instead of `.split(' ')` to allow for multispaces then?

skion: Also, this can now raise a `ValueError`, which I'm not sure is desired (though haven't checked in detail).

MattBlack85: dang you're right, too focused on python3. For the multispace I don't know, doesn't `1*SP` stays for 1 space?

skion: I know the feeling :)

Re `1*SP` is one or more spaces... unnatural indeed.

MattBlack85: aaaaa gotcha, fixing also that then! :)

MattBlack85: fixed both @skion

JonathanHuot:
Somehow Travis-CI didn't add a link to the build, maybe because it has been canceled before starting (5months ago, by who, why?..)

So I have restarted the build which is passing, and adding a link : https://travis-ci.org/oauthlib/oauthlib/builds/290372743

MattBlack85:
thanks @JonathanHuot :+1:

JonathanHuot:
Would you want to add a new unit test which test headers with multiple spaces between `Bearer` and `<token>` ?

As @skion pointed out, it's validated by the RFC.

Once done, the PR LGTM, ready for the merge! :-)

MattBlack85:
@JonathanHuot I added that header herehttps://github.com/MattBlack85/oauthlib/blob/38c56b20185ed84989563283f5855f82ef6d91ae/tests/oauth2/rfc6749/test_tokens.py#L68

wasn't this what @skion had in mind?

JonathanHuot:
I let him reply, however I understood that we need to accept several spaces `1*SP` after `Bearer`.

The line you mention is testing several spaces `1*SP` after `Beaver` (a wrong request, then), which is also good to have.

skion:
Yes, I think you're missing that positive test case, maybe make [this](https://github.com/MattBlack85/oauthlib/blob/38c56b20185ed84989563283f5855f82ef6d91ae/tests/oauth2/rfc6749/test_tokens.py#L62-L64) into a list like the fake headers and add a test with multiple spaces between Bearer and the token there?

MattBlack85:
@skion looking again at the PR it seems I already made that check here https://github.com/oauthlib/oauthlib/pull/491/files#diff-6096f6f429036633b8a4a4156878b1c4R72 with related test https://github.com/oauthlib/oauthlib/pull/491/files#diff-6096f6f429036633b8a4a4156878b1c4R110

JonathanHuot:
LGTM :-)

PR #388, opened 2015-08-31
Add support for HMAC-SHA256

mick88:

PR #498, opened 2017-10-27
Add support for HMAC-SHA256 (builds on PR#388)

ViktorHaag:
This pull incorporates changes from PR #388 as it appears orphaned.

I cleaned up the comments for the HMAC-SHA256 signing method, and added tests for the SHA1 and SHA256 signing methods (I verified the signature generation code against node's `oauth-1.0a` and `crypto` modules, to be satisfied that the test signature values were valid).

### DIFF
--- a/oauthlib/oauth1/rfc5849/signature.py
+++ b/oauthlib/oauth1/rfc5849/signature.py
@@ -478,20 +478,20 @@ def sign_hmac_sha256_with_client(base_string, client):
 
 
 def sign_hmac_sha256(base_string, client_secret, resource_owner_secret):
-    """**HMAC-SHA1**
+    """**HMAC-SHA256**
 
-    The "HMAC-SHA1" signature method uses the HMAC-SHA1 signature
-    algorithm as defined in `RFC2104`_::
+    The "HMAC-SHA256" signature method uses the HMAC-SHA256 signature
+    algorithm as defined in `RFC4634`_::
 
-        digest = HMAC-SHA1 (key, text)
+        digest = HMAC-SHA256 (key, text)
 
     Per `section 3.4.2`_ of the spec.
 
-    .. _`RFC2104`: http://tools.ietf.org/html/rfc2104
+    .. _`RFC4634`: http://tools.ietf.org/html/rfc4634
     .. _`section 3.4.2`: http://tools.ietf.org/html/rfc5849#section-3.4.2
     """
 
-    # The HMAC-SHA1 function variables are used in following way:
+    # The HMAC-SHA256 function variables are used in following way:
 
     # text is set to the value of the signature base string from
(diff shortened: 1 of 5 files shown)


oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -469,6 +469,63 @@ def sign_hmac_sha1(base_string, client_secret, resource_owner_secret):
...
+                          )
+
+
+def sign_hmac_sha256(base_string, client_secret, resource_owner_secret):
+    """**HMAC-SHA256**
+
+    The "HMAC-SHA256" signature method uses the HMAC-SHA256 signature
+    algorithm as defined in `RFC4634`_::
+
+        digest = HMAC-SHA256 (key, text)
```
lepture: There is no such thing in RFC.

lepture:
Actually there is a way to register your own signature method.
https://github.com/idan/oauthlib/blob/master/oauthlib/oauth1/rfc5849/__init__.py#L52

This **HMAC-SHA256** is not a part of the specification in RFC. This patch should not be included in this repo.

oauthlib/oauth1/rfc5849/signature.py
ViktorHaag: Agreed. Note that my comment didn't claim it was in RFC5849, but defined in RFC4634 (as one of the suite of accepted Secure Hash Algorithms beyond SHA-1).

I took 5849's provision of three signature methods as _not exclusive_. In section 3.4, it says "OAuth _does not mandate_ a particular signature method" (my emphasis), and so it seems useful, with little harm, to add support for SHA2-family signing methods.

If the maintainers would prefer to have clients add support for more modern signing methods via extension, rather than provided in the library, I can understand that decision. I would have thought having the support in-library would at least help clients ensure consistency, but it's clearly not my decision.

lepture: @ViktorHaag See my comment. There is a register function to extend signature methods. The folder is named as rfc5849, things beyond this scope should not be included.

cc @thedrow @ib-lundgren

ViktorHaag: Yes, I understand your point. I had assumed from the comments in #388, that the blocking factor here was a lack of testing and documentation, and not the idea in the first place. It was or that reason that I decided to help by "finishing" the original PR. If the maintainer group would prefer not to include it, that's fine: I'm glad there is a way to extend the package usefully.

thedrow:
I have given this some thought.
I don't see the harm in providing this implementation to our users.
It's a pretty common usecase.

thedrow:
Thank you @ViktorHaag @mick88 

@lepture If you strongly disagree let's chat about this over email.
We can always revert before making a new release.

PR #499, opened 2017-11-04
Add missing NotImplementedError

greyli:


skion:
This should not be breaking any existing clients, and if it is, it merely exposes a bug in the client code, which can only be a good thing.

LGTM :+1:

greyli:
It's my pleasure :)

PR #500, opened 2017-11-04
Check access token in self.token dict

greyli:
Since we can supply `access_token` in the `token` dict, so we need to check both places at here.


thedrow:
Thanks!

PR #504, opened 2017-11-23
Request object not given to confirm_redirect_uri

spack971:
It seems that the request object is never given to the `confirm_redirect_uri`.

```console
$ git grep -l confirm_redirect_uri oauthlib
oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
oauthlib/oauth2/rfc6749/request_validator.py
```

```console
$ grep -A 1 confirm_redirect_uri oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
        if not self.request_validator.confirm_redirect_uri(request.client_id, request.code,
                                                           request.redirect_uri, request.client):
```

This is an issue since I enclose the database session within the request object. That means that I cannot proceed with the query to fulfill the request.

### DIFF
--- a/examples/skeleton_oauth2_web_application_server.py
+++ b/examples/skeleton_oauth2_web_application_server.py
@@ -67,7 +67,7 @@ class SkeletonValidator(RequestValidator):
         # state and user to request.scopes and request.user.
         pass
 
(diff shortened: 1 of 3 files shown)


skion:
I see no reason not to do this: It aligns the validator signature with most of the others without a breaking change. :+1:

PR #506, opened 2017-12-12
Remove special treatment of nonstandard "expires" parameter 

seedifferently:
Hello,

Since the `expires` parameter is not part of the rfc6749 spec, it should be ignored per https://tools.ietf.org/html/rfc6749#section-5.1.

Furthermore, the assumption that `expires` is an alias of (and takes precedence over) `expires_in` breaks implementations that utilize an `expires` parameter for other purposes.

This MR removes handling of the nonstandard `expires` parameter altogether, ensuring that the `expires_in` parameter is properly utilized.


duaneking:
I'm 100% for this as it seems to violate the spec.

The one single question I have is "what does this break"?

duaneking:
I just validated that https://tools.ietf.org/html/rfc6750 contains no explicit "expires" field, only the expires_as in section 4, on page 10.

I just validated that https://tools.ietf.org/html/rfc6749 contains no explicit "expires" field, only the "expires_in" field.

From what RFC did this field come from?  Or was it just a dev trying to add support for a product with lazy developers that uses the wrong fields?

duaneking:
Just want to point out that the build isn't working; I try to check on its status with this and I get a "we cant find your build" page.  There may be a delay.

JonathanHuot:
Previous issues https://github.com/oauthlib/oauthlib/pull/268 shown that it was added for Facebook compliance, however, `requests-oauthlib` seems to have fixtures for various OAuth2.0 providers already, including this `expires` bug: https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/compliance_fixes/facebook.py#L25-L27

So, LGTM !

duaneking:
Ok so to confirm: looking deeper this, its a compat fix for Facebook Engineers being stupid?

duaneking:
Referencing #524 at https://github.com/oauthlib/oauthlib/issues/524 as a placeholder for the build issue.

The other issues can be worked on concurrently.

duaneking:
Sorry to post yet again in this thread,  but I wanted to update everybody that I created a new defect about this on Facebook's side with them directly,  and gave them the link to this issue so they are aware of the defect in their code.

We should depreciate the workaround, but I think its even more important to call it out as a workaround in code comments, explaining exactly why it exists and highlighting the violation of the RFC spec on the part of the provider.

skion:
Since this is not backward compatible, shall we tag it for 3.0.0?

duaneking:
I agree, good to have but lets make this a non-compat release.

skion:
Anyone here think this needs further work before merge?

PR #508, opened 2017-12-18
Added initial introspect support

JonathanHuot:
Suggested introspect implementation as discussed at https://github.com/idan/oauthlib/issues/497

### DIFF
--- a/oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
@@ -18,13 +18,14 @@ from ..grant_types import (AuthCodeGrantDispatcher, AuthorizationCodeGrant,
                            ResourceOwnerPasswordCredentialsGrant)
 from ..tokens import BearerToken
 from .authorization import AuthorizationEndpoint
+from .introspect import IntrospectEndpoint
 from .resource import ResourceEndpoint
 from .revocation import RevocationEndpoint
 from .token import TokenEndpoint
 
 
-class Server(AuthorizationEndpoint, TokenEndpoint, ResourceEndpoint,
-             RevocationEndpoint):
+class Server(AuthorizationEndpoint, IntrospectEndpoint, TokenEndpoint,
+             ResourceEndpoint, RevocationEndpoint):
 
     """An all-in-one endpoint featuring all four major grant types."""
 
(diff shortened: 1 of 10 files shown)


skion:
@JonathanHuot This looks neat! I'll can have a detailed look later...

oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
```diff
@@ -159,10 +162,11 @@ def __init__(self, request_validator, token_generator=None,
                                   token_types={'Bearer': bearer})
         RevocationEndpoint.__init__(self, request_validator,
                                     supported_token_types=['access_token'])
+        IntrospectEndpoint.__init__(self, request_validator)
```
skion: Is the asymmetry with the `RevocationEndpoint` in `supported_token_types` intentional here?

oauthlib/oauth2/rfc6749/request_validator.py
```diff
@@ -166,6 +166,26 @@ def is_within_original_scope(self, request_scopes, refresh_token, request, *args
...
+
+        The dict of claims is added to request.token after this method.
+
+        :param token: The token string.
+        :param token_type_hint: access_token or refresh_token.
+        :param request: The HTTP Request (oauthlib.common.Request)
+
+        Method is used by:
+            - Introspect Endpoint (all grants are compatible)
+        """
```
skion: Add a link to https://tools.ietf.org/html/rfc7662#section-2.2 here, and perhaps even mention the (currently) registered claims?

oauthlib/oauth2/rfc6749/request_validator.py
```diff
@@ -166,6 +166,26 @@ def is_within_original_scope(self, request_scopes, refresh_token, request, *args
         """
         return False
 
+    def introspect_token(self, token, token_type_hint, request, *args, **kwargs):
+        """Introspect an access or refresh token.
+
+        Called once introspect token request is validated. This method
+        should return a dictionary with any desired claims associated
+        with the *token*. The implementation can use *token_type_hint*
```
skion: You probably want to ensure people will return `None` for unknown tokens instead of raising some error, so that the endpoint return `{"active": false}` instead of an error code for unknown tokens as per spec. Maybe

> with the *token*, or `None` in case the token is unknown.

JonathanHuot: Agree with you

oauthlib/oauth2/rfc6749/request_validator.py
JonathanHuot: Good idea !

oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
JonathanHuot: I forgot it, thanks :-)

duaneking:
One nitpick I have is that this should be optional, and the ordering of the arguments constrictor would be a breaking change... would it be possible to make this new endpoint the last param with a disabled/None default so upstream users are not broken by it?

JonathanHuot:
Hi @duaneking, thanks for comments.

To better understand your concerns of ordering, could you give me an example which broke implementation ? My patch doesn't update constructor nor arguments of it. However it adds a parent class which is ordered, but I can't imagine an example where it breaks.

Also, about the optional thing, at the moment no HTTP endpoint is associated with `create_introspect_response`, so the request validator's `introspect_request` will never be called. So I don't think it breaks any upstream users. However, please don't hesitate to correct me with an example to help me to understand.

Finally, I mimick what have been done for the **revocation** RFC, which has been added to a `Server` but is not active as long as we don't declare `create_revocation_response` HTTP endpoint.

I appreciate your time,
Thanks

JonathanHuot:
Any feedback ?

skion:
I don't spot any incompatibilities either, and would be happy if we include this in our next release. :+1:

PR #514, opened 2018-02-06
Update repository location in Travis

skion:
We'll need to configure Travis as well still...


JonathanHuot:
README.md should reflect the new Travis URL as well, (coveralls.io too)

Issue #511, opened 2018-01-28
Migration to oauthlib community


JonathanHuot:
Hi everyone,

Since @idan accepted the oauthlib community migration, as a team, we should list what do we need to progress as a true community. I suggest to start with a small list, and please anyone, feel free to participate by adding any suggestions :-) 

Define/Improve release process:
- [x] : **Tagging**
Wrong releases numbers: we have **2.0.3** on `github's releases`, **2.0.5** in `__init__.py` and **2.0.6** on `pypi`...something wrong for sure
- [x] : **Publishing**
I'd recommand to continue to use Travis for publishing, however we could define version directly by using environment variable `TRAVIS_TAG` (see example at https://github.com/thomsonreuters/bottle-oauthlib/blob/master/setup.py instead of our current hard-coded value: https://github.com/oauthlib/oauthlib/blob/master/oauthlib/__init__.py ). Also, I've seen @ib-lundgren is the actual publisher of the pypi package, he looks inactive since years, but dunno if it's an issue.  
- [x] : **Documentation**
README needs to be updated with the updated badge for **travis-ci.org repository** URL (owner must enable it, AFAIK), and also add a badge for the **documentation build**, cuz it's failing since long time: https://readthedocs.org/projects/oauthlib/builds/6483131/
**DONE IN**: https://github.com/oauthlib/oauthlib/pull/520 
- [x] : **Communication**
No objection to continue using github issues and [Google+](https://plus.google.com/communities/101889017375384052571), however it seems a bit outdated. Also, the #oauthlib IRC channel is empty, ~wondering if we could create a Slack room if interested?~
**CONCLUSION**: gitter came to the rescue. Feel free to join at https://gitter.im/oauthlib/Lobby

Future, roadmap:
- [ ] : Do periodic bugs washes could be beneficial

Also, a quick round table could be great. I start the introductions, I'm currently working on a OAuth2.0 RequestValidator implementation with bottle, and I've never worked with OAuth1.0, nor Django nor Pyramid nor Flask. However, I'm trying to have a good knowledge around the RFC involved here (oauth2, introspect, revocation, jwt...). I have not started yet the OpenID integration, but it will come soon.

skion:
All great ideas @JonathanHuot!

And what do people think about:

- Cleaning up some of the old branches.
- Starting to use GitHub's milestones to plan releases (happy to make a first run on e.g. 3.0, 3.1, 4.0).
[...]

PR #520, opened 2018-03-12
Add shields for Python versions, license and RTD

skion:
This makes the documentation accessible from a shield, and two other shields.


JonathanHuot:
LGTM, related to https://github.com/oauthlib/oauthlib/issues/511

Issue #512, opened 2018-02-02
OAuthLib upcoming release 3.0.0


wiliamsouza:
**This is a backward incompatible release**

Release notes for OAuthLib 3 to be released in 2018-02-05.

Updates to changelog:

3.0.0 Add new release changelog
2.0.5 Was released in 2017-10-18 according https://pypi.python.org/pypi/oauthlib/2.0.5
2.0.5 Add an entry explaining the problem of releasing backward incompatible version.
2.0.6 This is revert release from  backward incompatible added in 2.0.5 notice.

CHANGELOG.rst
=

3.0.0 (2018-02-05)
------------------

**This is a backward incompatible release**

This release improves OpenID Connect supporting removing the need of using `grant_type=openid` in token endpoint and when defining a application credential the `authorization_grant_type` can be same already used (`authorization-code`) both for OpenID Connect and OAuth2.

Replaces the way to send a token from `Authorization: JWT eyJhbGciOiAiUlMyNTYifQ...` to `Authorization: Bearer eyJhbGciOiAiUlMyNTYifQ...`.

* OpenID connect improvements #484.
* Openid connect jwt #488

2.0.6 (2017-10-19)
------------------

This is revert release from  backward incompatible added in 2.0.5

* Fix cliend_id in web request body #505
* Added bottle-oauthlib #509
* Add support for HMAC-SHA256 (builds on PR#388) #498
* Check access token in self.token dict #500
* When deploying a release to PyPI, include the wheel distribution #496
* Include license file in the generated wheel package #494
* Refactor OAuth2Error
* Fix OAuth2Error.response_mode for #463.
* Documentation improvement.
* initialization of variable #487
* Pass through nonce in code flow #481

2.0.5 (2017-10-18)
------------------
**Do not use this version**

By mistake backward incompatible changes were released in this version the next reverts it.

wiliamsouza:
@lepture @wwidnyana @evonove @masci @lundgren @lukasa As stated by OAuthLib release process [1] i'm letting you know before hand about this upcoming release!

[1] https://github.com/oauthlib/oauthlib/blob/master/docs/release_process.rst

duaneking:
We really should support the official RFC for JWT before we claim to support OpenIDC, as we don't right now, and JWT is a hard requirement for it.

wiliamsouza:
Can you please enumerate those RFCs?
[...]

PR #522, opened 2018-03-12
Fixed "make" command to test upstream with local oauthlib.

JonathanHuot:
Doing `make` command to test upstream compatibility is great, however it has different flaws I fixed:
1) remove oauthlib references in upstream's project and replace it with local "dev" oauthlib: every dev should be able to run `make` to test their local changes before push.
2) clean rule: cleaning the working directories should not be done automatically, else it becomes a pain to debug them
3) python versions list should be based on the upstream's compatibility, and not on the oauthlib's compatibility (e.g. running plain `tox` everytime we can).  
4) added `bottle-oauthlib`

TBC: wait for https://github.com/requests/requests-oauthlib/pull/306 to be integrated to master & we will able to replace `Travis` build command from `tox` to `make`. However, if one of their builds failed for a different reason than us, it breaks our CI job.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,47 +1,57 @@
-PYS = py27,py34,pypy
+# Downstream tests (Don't be evil)
+#
+# Try and not break the libraries below by running their tests too.
+#
+# Unfortunately there is no neat way to run downstream tests AFAIK
+# Until we have a proper downstream testing system we will
+# stick to this Makefile.
+#---------------------------
+# HOW TO ADD NEW DOWNSTREAM LIBRARIES
+#
+# Please specify your library as well as primary contacts.
+# Since these contacts will be addressed with Github mentions they
+# need to be Github users (for now)(sorry Bitbucket).
+#
+clean:
+	rm -rf .tox
+	rm -rf bottle-oauthlib
+	rm -rf django-oauth-toolkit
+	rm -rf flask-oauthlib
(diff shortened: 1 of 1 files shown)


JonathanHuot:
FYI,
With this Makefile, we have 
- release 2.0.7 https://github.com/skion/oauthlib/tree/2.0.7 working as expected (except for requests-oauthlib in py33, as mentioned above).
- master (3.0.0) failed as expected, discussion is in progress about incompatibility here https://github.com/oauthlib/oauthlib/issues/512

Makefile
```diff
@@ -1,47 +1,57 @@
...
+django:
 	#---------------------------
 	# Library: evonove/django-oauth-toolkit
 	# Contacts: evonove,masci
 	# (note: has tox.ini already)
-	git clone https://github.com/evonove/django-oauth-toolkit.git
-	cd django-oauth-toolkit && tox -e "$(PYS)"
-	rm -rf django-oauth-toolkit
+	cd django-oauth-toolkit 2>/dev/null || git clone https://github.com/evonove/django-oauth-toolkit.git
+	cd django-oauth-toolkit && sed -i.old 's,deps =,deps= --editable=file://{toxinidir}/../,' tox.ini && tox -e py27,py35,py36
```
skion: This one appears to need an extra newline for the `docs` build...

PR #523, opened 2018-03-13
Updated Travis releases part

JonathanHuot:
Since the migration to the community https://github.com/oauthlib/oauthlib/issues/511, we are deprecating IRC in favor of Gitter, and we need to provide a Github Releases tab in sync with pypi and our release schedule.

IIUC, the ChangeLog will not be done automatically by Travis, but at least we should have the correct assets (as opposite of https://github.com/oauthlib/oauthlib/issues/511 / Tagging issue). This step could always be done manually, after the release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,12 @@ script: tox
 
 after_success: coveralls
 notifications:
-  irc: irc.freenode.org#oauthlib
+  webhooks:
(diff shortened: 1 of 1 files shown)


skion:
Nice one @JonathanHuot! :1st_place_medal:

PR #530, opened 2018-03-19
Merge 2.0.7 back to master

skion:
I've cherry picked two commits from 2.x to update the changelog in master.

Alternatively, if we're planning to do further 2.x.y releases we could still revert the offending (breaking) commits from master so that we don't need the 2.x branch in the first place.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,23 @@
 Changelog
 =========
 
+2.0.7 (2018-03-09)
+------------------
+
+* Moved oauthlib into new organization on GitHub.
+* Include license file in the generated wheel package. (#494)
+* When deploying a release to PyPI, include the wheel distribution. (#496)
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,29 @@
...
+* Add shields for Python versions, license and RTD. (#520)
+* Fix ReadTheDocs build (#521).
+* Fixed "make" command to test upstream with local oauthlib. (#522)
+* Replace IRC notification with Gitter Hook. (#523)
+* Added Github Releases deploy provider. (#523)
+
+2.0.6 (2017-10-20)
+------------------
+
+* 2.0.5 contains breaking changes.
```
JonathanHuot: should add "Fix 2.0.5 release", the current 2.0.6 changelog is a bit confusing

PR #542, opened 2018-05-01
Avoid populating spurious token credentials

skion:
This patch ensures we do not pick up client attributes too early.

### DIFF
--- a/oauthlib/oauth2/rfc6749/clients/base.py
+++ b/oauthlib/oauth2/rfc6749/clients/base.py
@@ -111,6 +111,7 @@ class Client(object):
         self.state_generator = state_generator
         self.state = state
         self.redirect_url = redirect_url
+        self.code = None
         self.expires_in = None
         self._expires_at = None
         self._populate_token_attributes(self.token)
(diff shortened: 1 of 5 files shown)


oauthlib/oauth2/rfc6749/clients/base.py
```diff
@@ -459,8 +460,14 @@ def _add_mac_token(self, uri, http_method='GET', body=None,
                                             hash_algorithm=self.mac_algorithm, **kwargs)
         return uri, headers, body
 
-    def _populate_attributes(self, response):
-        """Add commonly used values such as access_token to self."""
+    def _populate_code_attributes(self, response):
+        """Add attributes from an auth code response to self."""
+
+        if 'code' in response:
+            self.code = response.get('code')
```
wiliamsouza: Initialize `self.code` in `__init__`?

oauthlib/oauth2/rfc6749/clients/mobile_application.py
```diff
@@ -168,5 +168,5 @@ def parse_request_uri_response(self, uri, state=None, scope=None):
         .. _`Section 3.3`: https://tools.ietf.org/html/rfc6749#section-3.3
         """
         self.token = parse_implicit_response(uri, state=state, scope=scope)
-        self._populate_attributes(self.token)
+        self._populate_token_attributes(self.token)
```
wiliamsouza: Is there test a `populate_attributes` for mobile application?

skion: There is now! :+1:

PR #545, opened 2018-05-08
Release 2.0.8

skion:
This patch release addresses the following...

* Fixed some copy and paste typos (#535)
* Use secrets module in Python 3.6 and later (#533)
* Add request argument to confirm_redirect_uri (#504)
* Avoid populating spurious token credentials (#542)

We should discuss if we want #525 in here as well...

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -28,3 +28,4 @@ Joel Stevenson
 Brendan McCollam
 Jonathan Huot
 Pieter Ennes
+Olaf Conradi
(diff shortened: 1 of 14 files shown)


wiliamsouza:
+1 to release #525 as part of 2.0.8

skion:
@wiliamsouza I looked at #525 again and actually that would break any existing OIDC implementation just because the files have moved around. So I would propose it goes in 3.0.0 instead of 2.0.8 because of that.

skion:
The Makefile tests pass on this now...

skion:
@thedrow Makes sense :+1: 

Closing in favour of #548.

Issue #264, opened 2014-09-15
Invalid HTTP statuses on error response


prudnikov:
RFC 6749 Section 5.2 (http://tools.ietf.org/html/rfc6749#section-5.2) describes error response. It says that server should respond with HTTP 400 (Bad request). But oauthlib uses HTTP 401 (Unauthorized) instead for errors such as `InvalidClientError`, `InvalidGrantError`, `UnauthorizedClientError` and other related errors.

For example https://github.com/idan/oauthlib/blob/master/oauthlib/oauth2/rfc6749/errors.py#L181

samskiter:
@ib-lundgren Hi - just stumbled into this very issue and would like to chime in. From reading the spec and looking around other answers, IMO a 400 is the correct response.

Firstly, the spec:

>The authorization server responds with an HTTP 400 (Bad Request)
   status code (unless specified otherwise) and includes the following
   parameters with the response:

There's no MUST in the 400 requirements, but there's no MAY either.

Secondly, and the reason I stumbled into this was through a bug of sorts in the sun JDK on my client (see here: http://stackoverflow.com/a/7524681/1137254 )

Essentially, my client's HTTP stack would close the connection early for a 401, but the OAuth stack above (google-http-oauth-client) would attempt to continue reading and parse the results. I think this was written into OpenJDK per the definition and expectations around a 401. IMO OpenJDK is wrong too and I think I'd hit this same issue if I were to get an "invalid_client" 401 during /token endpoint requests.

jmbowman:
Adding another vote for 400 being the correct error code here.  We used to use django-oauth2-provider for oauth2 authentication from mobile apps, and it returned 400 in this case.  When switching to django-oauth-toolkit (which uses oauthlib) in order to upgrade to newer Django versions, we had to hack the token request view to return a 400 instead to keep the mobile apps working correctly.

Here's one of the authors of the spec specifically stating that a 400 is the intended response for invalid authentication credentials when requesting a token, and why: https://www.ietf.org/mail-archive/web/oauth/current/msg02127.html

And here's [another example](https://github.com/seegno/angular-oauth2/issues/36) of a project that's having trouble coping with oauthlib using a 401 in these circumstances.

LennyLip:
+1 for 400

rinatio:
Vote for 400

JonathanHuot:
I'm reviving this old thread & start a PR with the changes suggested above.

JonathanHuot:
[...]

PR #561, opened 2018-07-04
`invalid_grant` status code should be 400

freeduerinckx:
According to section 5.2 of rfc 6749
(https://tools.ietf.org/html/rfc6749#section-5.2)

A server should respond with 400 in case of an invalid grant. The
given grant is invalid and the client should give other data.

A 401 is not applicable here because the client is required to give
suitable Authorization header field which doesn't make any sense if
you are trying to acquire a grant authentication.

According to sections 10.4.1 and 10.4.2 of rfc 2616
(https://tools.ietf.org/html/rfc2616#section-10.4.1)

WIP: these changes breaks downstream libs tests:
- django
- flask


JonathanHuot:
Thanks for the PR! This issue was highlighted at the beginning of https://github.com/oauthlib/oauthlib/issues/264 but no PR has been proposed since. Few more 401 must be 400 also as mentionned in the #264 .

We should release that as part of the 3.0.0 branch.

jvanasco:
The docstring on the error should reference the RFC sections mentioned above.

PR #565, opened 2018-07-21
Fix OIDC tests

skion:
After merging #525 we noticed a drop in test coverage from 96% to 92%.

This PR attempts to correct that, re-enabling the undiscovered tests by adding two `__init__.py` files, and fixing a number of import problems.

### DIFF
--- a/oauthlib/openid/connect/core/endpoints/pre_configured.py
+++ b/oauthlib/openid/connect/core/endpoints/pre_configured.py
@@ -8,29 +8,31 @@ for providing OpenID Connect servers.
 """
 from __future__ import absolute_import, unicode_literals
 
-from ..grant_types import (
+from oauthlib.oauth2.rfc6749.endpoints import (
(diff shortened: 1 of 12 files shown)


skion:
@wiliamsouza Can you check if this is OK?

JonathanHuot:
Also, it raises an interesting question that we don't have the actual details of the tests runs by `nosetests` or `Travis`. I'm suggesting to add `-v` (verbose) to `nosetests` to trace records of each unittest runs into the Travis console.

I have also tested to produce an html report with `nose-htmloutput` and it's clear that each `tests/openid` tests are correctly run in this branch.

Do you see guys, an interest of adding the html report automatically when running tox ?

skion:
@thedrow Rebased

JonathanHuot:
Wonderful achievement here, congrats to @wiliamsouza and @skion to make this a reality !

Issue #539, opened 2018-04-29
Community Lacks A Code of Merit


duaneking:
Per the email thread where I took a PHD to task for not using the scientific method and being intellectually lazy, this community lacks a set of community contribution merit standards and as a result people who wish to contribute have no official guide on what merits their work have.

I have personally worked on projects without a merit standard , as well as ones where they explicitly list out merits. 

Based on the feedback form that email thread, I suspect that we should put more effort into ours.  I would prefer personally we simply start with http://code-of-merit.org/ as a base and edit as needed, but I wanted to open the floor to the rest of the team.

JonathanHuot:
I never worked on projects **with** an explicit merit standard, I'm curious to know your feedback on this ? 

I've read the code-of-merit above, some of the bullets points makes sense and I think it could helps on making progress on the project, though.

duaneking:
It really does help with egos, as well as direct efforts to what really should be the real goal of working, usable software.

Issue #540, opened 2018-04-29
Community Lacks a Code of Conduct


duaneking:
Per the team email thread where I took a PHD to task for not using the scientific method and being intellectually lazy, this community lacks a set of community contribution merit standards and as a result people who wish to contribute have no official guide on what conduct is preferred or allowed for this project.

I have personally worked on projects without a code of conduct,  as well as ones where they explicitly list out everything in detail. 

Based on the feedback form that email thread, I suspect that we should put more effort into ours.

I would prefer personally we simply start with https://github.com/domgetter/NCoC  as a base and edit as needed to provide protections against the very worst possibilities the internet has to offer if the team does not feel they are already covered by it,, but I wanted to open the floor to the rest of the team.

JonathanHuot:
I read the NCoC for the first time, it's funny while being pragmatic. 

As an alternative, I share another widely used code-of-conduct, short and clear in few bullet points. 
https://www.contributor-covenant.org/version/1/4/code-of-conduct.html

skion:
I'm fine with the Contributer Convenant, but would also like to suggest the [Django CoC](https://www.djangoproject.com/conduct/), which was praised [here](https://opensource.guide/code-of-conduct/).

I'd not be in favour of the NCoC, as I find the language in it unnecessarily stimulating while it doesn't really add anything to "common sense", or really having no CoC.

JonathanHuot:
I'm also find with Django CoC, it could flag our project more "pythonic" than for the others CoC.

Others?

duaneking:
The Django CoC is a good start but it doesn't clearly differentiate between a misunderstanding, bad communication, or simple lack of context.

The one thing I see a lot of lately in different communities is that somebody will have done something years ago, they get called out on it now, and then they are ruined by who they USED to be despite sincere and honest efforts to be a better person.  We see that with Hollywood and other places a lot right now.. and frankly this internet lynchmob BS is counter productive so I dont want it here.

People should be free to grow up.  They should be free to act like adults, free to improve themselves, be free to be a better person then they used to be, without stuff they do OUTSIDE the project being an issue for them.  I say judge people by how they act here, not what they posted on twitter 10 years ago.

JonathanHuot:
Since it relates to the oauthlib community, it would be good to have a thumbs up, or any opinions from others? Poke @lepture, @thedrow, @ViktorHaag, @wiliamsouza, @idan 

[...]

PR #573, opened 2018-08-12
Add NCoC and Code of merit

JonathanHuot:
Should fix https://github.com/oauthlib/oauthlib/issues/539 and https://github.com/oauthlib/oauthlib/issues/540

The discussion is still open for everyone.


duaneking:
@JonathanHuot  I liked the start but not the additions. Looking at the current version, it includes things that hurt peoples ability to contribute. I think using the django version is a mistake.

PR #583, opened 2018-08-30
Mention `oauth_body_hash` in OAuth1 client docs

ngoue:
While the previous documentation was not wrong in that non-formencoded data are not included in the signature for traditional OAuth1 service providers, the library does still include an `oauth_body_hash` for non-formencoded data. 

Update the documentation to include mention of the `oauth_body_hash` with a notice that validation of said parameter may not be supported by all service providers, but will nevertheless provide an additional integrity check for those that do support it.


ngoue:
I'm not sure if it's worth mentioning, but I'm not sure whether the `oauth_body_hash` ever made it past draft stage:

https://tools.ietf.org/html/draft-eaton-oauth-bodyhash-00

I figured since it's already implemented, it would be helpful to include mention in the docs.  We needed  to include it in our requests and it wasn't readily apparent in the docs that this was available.

JonathanHuot:
LGTM, however I'm not a OAuth1 expert :-) anyone @ib-lundgren @idan ? 
Thanks for your contribution !

Issue #406, opened 2016-02-22
scope is optional in code response_type?


ryarnyah:
Hi!

When i read the doc in ***oauthlib/oauth2/rfc6749/grant_types/authorization_code.py***, i can see :
`        scope
                OPTIONAL.  The scope of the access request as described by
                `Section 3.3`_.`

But this check assume that scopes is mandatory
`
            if not request.scopes:
                raise ValueError('Scopes must be set on post auth.')`

Am i misunderstanding something?

btimby:
This one bit me too. I monkey-patched it.

```python
# A bit of monkey patching here. OAuthlib states that scope are optional, also
# RFC6749 confirms this, however, one method of the AuthorizationCodeGrant
# class requires scopes to be defined. As I don't know if we are going to use
# scopes, or what the scopes might be, I am going to monkey-patch this class to
# make scopes optional.
#
# https://github.com/idan/oauthlib/issues/406

# Some imports we need for the patching of the monkey...
from oauthlib.oauth2.rfc6749 import errors


def create_authorization_response(self, request, token_handler):
    """
    Monkey-patched version of this method that allows undefined scopes.
    """
    try:
        # Right here is where the base method checks scopes. We omit this check
        # but the rest of the method body is identical.

        self.validate_authorization_request(request)

        log.debug('Pre resource owner authorization validation ok for %r.',
                  request)
    except errors.FataClientError as e:
        log.debug('Fatal client error during validation of %r. %r.',
                  request, e)
        raise
    except errors.OAuth2Error as e:
        log.debug('Client error during validation of %r. %r.', request, e)
        request.redirect_uri = request.redirect_uri or self.error_uri
        return {
            'Location': common.add_params_to_uri(request.redirect_uri,
                                                 e.twotuples)
        }, None, 302

    grant = self.create_authorization_code(request)
    for modifier in self._code_modifiers:
        grant = modifier(grant, token_handler, request)
    log.debug('Saving grant %r for %r.', grant, request)
    self.request_validator.save_authorization_code(
        request.client_id, grant, request)
    return self.prepare_authorization_response(
        request, grant, {}, None, 302)


from oauthlib.oauth2.rfc6749.grant_types.authorization_code import (
    AuthorizationCodeGrant, log
)

AuthorizationCodeGrant.create_authorization_response = \
    create_authorization_response

# Now we can import the rest of what we need from oauthlib.
```

JonathanHuot:
[...]

PR #584, opened 2018-09-02
Make scope optional for authorization code grant

luhn:
As discussed in #406.  Sibling PR to #475.

Currently a WIP.  I'm having some trouble perhaps someone can help with:  Remove the scopes check causes [this test](https://github.com/oauthlib/oauthlib/blob/master/tests/oauth2/rfc6749/endpoints/test_base_endpoint.py#L23) to fail, because the test was using it as an example of a server error.  However, I can't find another error raised that isn't an OAuth2Error or FatalClientError.

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py
@@ -140,7 +140,6 @@ class AuthorizationCodeGrant(GrantTypeBase):
                               oauthlib.oauth2.BearerToken.
         :returns: headers, body, status
         :raises: FatalClientError on invalid redirect URI or client id.
-                 ValueError if scopes are not set on the request object.
 
         A few examples::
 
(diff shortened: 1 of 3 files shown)


luhn:
Alright, I was able to rework it to hit https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py#L401

JonathanHuot:
Hi @luhn, thanks for this PR !

Do we have an unit test testing specifically this ? (auth + no-scopes) We should take this PR opportunity to add it.

skion:
Scope *is* a [required parameter](http://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) for OIDC. Can we use this PR to make that explicit, if not already so? Or is the change in this PR unrelated...

luhn:
I would vote the change is unrelated, as the `openid` scope is required by OIDC, which is not enforced before or after this PR.

Issue #578, opened 2018-08-21
inline docs missing param?


jvanasco:
https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/request_validator.py#L67-L83

shouldn't this read:

        :param client_id: Unicode client identifier
        :param request: oauthlib.common.Request

not just

        :param request: oauthlib.common.Request

?

JonathanHuot:
Hi @jvanasco, yes it should be!
Could you open a PR with this change please ?

Thanks a lot to help improve the doc

jvanasco:
yeah. i think this is in a few places, so I'll do a pass and make a consolidated pr.

jvanasco:
This was addressed in #588

PR #579, opened 2018-08-27
initial docstring fixes

jvanasco:
this is an initial attempt at #578

I reviewed the oauth2 docs for missing or out-of-order docstrings in general, and fixed what i could based on my familiarity with the library and existing strings for the param in other functions.

the following are still deficient. if someone can offer suggestions for them, I will integrate them into this PR>

# parameters.py

* `prepare_grant_uri` there is no docstring for the `uri` param. 
* `prepare_token_revocation_request` missing docstring for `uri` and `callback`
* `parse_implicit_response` missing docstring for `uri`, `state` and `scope`

# tokens.py

* `prepare_mac_header` there is no docstring for `token`, `nonce`, `body`, `ext`
* `prepare_bearer_uri` no docstrings for params
* `prepare_bearer_headers` no docstrings for params
* `prepare_bearer_body` no docstrings for params
* `BearerToken` no methods have docstrings for params

# grant_types/base

* `GrantTypeBase.add_token`  what is `token`?
* `GrantTypeBase.prepare_authorization_response`  what is `token`? headers, body, status

PR #588, opened 2018-09-10
docstrings fix, try #2

jvanasco:
This is a second version of #579


oauthlib/oauth2/rfc6749/grant_types/base.py
```diff
@@ -116,14 +116,32 @@ def register_code_modifier(self, modifier):
...
+        :param token_handler: A token handler instace, for example of type
+                              oauthlib.oauth2.BearerToken.
+        """
         raise NotImplementedError('Subclasses must implement this method.')
 
     def create_token_response(self, request, token_handler):
+        """
+        :param request: OAuthlib request.
+        :type request: oauthlib.common.Request
+        :param token_handler: A token handler instace, for example of type
```
JonathanHuot: Typo error `instace` i/o `instance` in all token_handler locations :-)

oauthlib/oauth2/rfc6749/parameters.py
```diff
@@ -133,15 +132,19 @@ def prepare_token_revocation_request(url, token, token_type_hint="access_token",
...
-    token_type_hint  OPTIONAL.  A hint about the type of the token submitted
-    for revocation.  Clients MAY pass this parameter in order to help the
-    authorization server to optimize the token lookup.  If the server is unable
-    to locate the token using the given hint, it MUST extend its search across
-    all of its supported token types.  An authorization server MAY ignore this
-    parameter, particularly if it is able to detect the token type
-    automatically.  This specification defines two such values:
+    :param token: REQUIRED.  The token that the client wants to get revoked.
+
+    param:token_type_hint: OPTIONAL.  A hint about the type of the token
```
JonathanHuot: `param:`  instead of `:param `

oauthlib/oauth2/rfc6749/grant_types/base.py
jvanasco: ack! sorry! dealing with the merge-commit messages was so messy, so I just redid all the work.

jvanasco:
I fixed those two. I also updated the formatting on the `errors` file, and updated anything that uses the format "see methodname" to be "see ``.methodname``", which (I believe) should render better on automatic doc generators.

Since there wasn't an official styleguide, I adhered to the following existing conventions as I standardized stuff.  None of these are of my own design, they already existed but weren't applied universally. I just tried to make things consistent.

1. params in the docs are listed in their order as args/kwargs
2. if there is a textual docstring, use a single blank line between the docstring and the params.
3. I put referenced class functions in backticks as I found them. (likely many more)
4. Every `:param foo:` line is considered a sentence and ends in a period. Every `:type bar:` line is considered an object reference and does not end in a period.

Finally,  if i didn't know what the param's doc was, I still included it but left it empty -- for someone to fill out later.  You can just regex for ":param (\w+):\n" to find them.

JonathanHuot:
Could you add these instructions in the contribution guide documentation part itself ?

I think it does not harm to have this clearly stated in our documentation. That's very good to list the changes you made and that would be a shame to lose it.

jvanasco:
absolutely. i'll push an update asap.

jvanasco:
This PR closes #578

Issue #585, opened 2018-09-07
Client web application does no longer send client_id


JonathanHuot:
I found a regression in master when used with requests/requests-oauthlib since  https://github.com/oauthlib/oauthlib/issues/495 has been merged. It's related to authorization grant/web application only.

Basic usage of requests-oauthlib is :

    sess = OAuth2Session(client_id)
    token = sess.fetch_token(token_url, client_secret=client_secret, authorization_response=request.url)

However, since the changes, `client_id` of the session is ignored. I think https://github.com/oauthlib/oauthlib/pull/505 fixed an use-case but broke another one. We should find a win-win solution.

requests-oauthlib code call at https://github.com/requests/requests-oauthlib/blob/master/requests_oauthlib/oauth2_session.py#L196-L198 and oauthlib issue here 
https://github.com/oauthlib/oauthlib/blame/master/oauthlib/oauth2/rfc6749/clients/web_application.py#L128.

JonathanHuot:
My first thinking is to revert back the changes in `prepare_request_body` to, by default, use the `self.client_id` set in the `WebApplicationClient` constructor.
Then, inline docs should be changed in adequation to add `&client_id=xx` to the `prepare_request_body` output.

Finally, to replace the original fix, I'll suggest to remove `client_id` from the args, and add a new argument to `prepare_request_body` like `include_client=True/False` to both add `client_id` and `client_secret` in the body, or not include both of them.

Thoughts?

JonathanHuot:
poke @Diaoul @skion @thedrow 

What about: 

> I'll suggest to remove client_id from the args, and add a new argument to prepare_request_body like include_client=True/False to both add client_id and client_secret in the body, or not include both of them.

Thanks

jvanasco:
I actually triggered this same issue in one of my tests, but filed it against requests/oauthlib here:  https://github.com/requests/requests-oauthlib/issues/330

I believe the issue is actually the fault of requests_oauthlib.  Their docs - actually the first example in their entire docs when you load the page - support specifying the `client_id` in the `OAuth2Session` constructor.  The logic at line 200 pulls the `client_id` from the kwargs, but does not have a fallback to pull it from the already constructed `WebApplicationClient` instance.

JonathanHuot:
[...]

PR #593, opened 2018-09-13
PR for #585, `client_id` behavior with `prepare_request_body`

jvanasco:
This PR addresses #585 

I had to make the new test a little bit 'dumb' to keep it simple while maintaining support under Python2 & Python3 

1. The order of arguments inserted into the request body is deterministic and can shift on different platforms (or even the same platform on different computers).  Instead of sorting the result in the test, or generating a sorted Response body, I just test against 2 possible permutations.

2. The test checks for a ValueError Exception and a DeprecationWarning. I couldn't manage to support a check for the expected error message on both platforms while maintaining a clear and concise test -- so I just omitted that check.


jvanasco:
Cross reference to paired requests-oauthlib PR - https://github.com/requests/requests-oauthlib/pull/331

jvanasco:
> However, the include_client_id stuff should be moved to parameters.prepare_token_request instead. 

Agreed. I'll make the change and circle back here. I was just updating the tests, and they were starting to lead me to this same conclusion.

jvanasco:
My latest commit has the following:

### `prepare_token_request`

This now accepts `include_client_id` and has the logic for popping the `client_id` out of the kwargs.  ** The default is `True`, thought it might be better as `None`. **

I updated the docstring to note some fields with an asterisk `*` are not explicit kwargs but have special meanings or are expected. I did the same with the ServiceApplication as the same confusing phenomena happened there -- i.e. a docstring for a kwarg that isn't an enumerated kwarg.

### `Client`.`prepare_request_body`

These were all extended to have an `include_client_id` argument as well. This is a pass-through to `prepare_token_request`; only the `WebApplicationClient` defaults to `True`. Others all default to `None` for backwards compatibility.  **This might not be the correct default for the other classes.**

They all implement this code **before** invoking `prepare_token_request`.

    kwargs['client_id'] = self.client_id
    kwargs['include_client_id'] = include_client_id

Only `WebApplicationClient` has some logic to compare an explicit `client_id` to `self.client_id` and the deprecation warning, because this is the only class I know of which has some docs/examples supporting/old-code using `client_id` as a keyword argument.

oauthlib/oauth2/rfc6749/parameters.py
```diff
@@ -118,10 +137,21 @@ def prepare_token_request(grant_type, body='', **kwargs):
     if 'scope' in kwargs:
         kwargs['scope'] = list_to_scope(kwargs['scope'])
 
+    # pull the `client_id` out of the kwargs. 
+    client_id = kwargs.pop('client_id', None)
+    if include_client_id:
+        if client_id is not None:
```
skion: It seems that `client_id` isn't used in this function except in this line... I'm not sure if I understand why we'd need `include_client_id` in this function's signature in that case.

skion: It feels a bit fiddly with the `None`, `False`, `True` options and I'd be inclined to put the logic in the individual client classes, but am happy with this too.

oauthlib/oauth2/rfc6749/parameters.py
```diff
@@ -118,10 +137,21 @@ def prepare_token_request(grant_type, body='', **kwargs):
...
+            params.append((unicode_type('client_id'), client_id))
+
     for k in kwargs:
+        # this handles: `code`, `redirect_uri`, or undocumented params
         if kwargs[k]:
             params.append((unicode_type(k), kwargs[k]))
 
+    if ('client_secret' in kwargs) and ('client_secret' not in params):
```
skion: This might deserve a comment...

tests/oauth2/rfc6749/clients/test_legacy_application.py
```diff
@@ -10,19 +10,26 @@
 
 from ....unittest import TestCase
 
+# this is the same import method used in oauthlib/oauth2/rfc6749/parameters.py
```
skion: Feels superfluous as a comment?

oauthlib/oauth2/rfc6749/parameters.py
jvanasco: good idea.

jvanasco: wow, thanks for this comment. i actually found a weird issue from this and am rewriting the function a bit!

jvanasco: fixed, and with a comment.

the outdated code will *never* find "client_secret" in params, because params is a list of tuples. 

    ('client_secret' not in params)

the search should have been something like this:

    ('client_secret' not in [p[0] for p in params])

but that's ugly and inefficient.

in the fix, I just popped client_secret from the kwargs and compared it to `None`.

JonathanHuot:
@jvanasco, I have no objections about the docstring changes. I bet we should do something better than playing with kwargs however that's not the purpose of this PR and what you did is great.

For `client_secret`,  it is odd that unittest didn't caught this ? maybe we could add this test. However that's the right change 👍

jvanasco:
Existing tests didn't catch an error, because the block always generates the right output - it just generates the right output in the wrong way.  The block correctly only acted on the `client_secret` if it was an empty string, but ran the comparison on every `client_secret` submission -- missing the shortcircuit exit that ensured it wasn't already in the kwargs.

    # this line will always evaluate to True if `client_secret` is in the kwargs, it should have exited if already present
    if ('client_secret' in kwargs) and ('client_secret' not in params):	
        # this comparison still worked correctly. 
        if kwargs['client_secret'] == '':

I'm not a fan of playing with the kwargs either, unfortunately it's just done a lot in this library. I did upgrade the docs whenever there is some kwarg action going on to make things more apparent.

JonathanHuot:
That's merged! Thanks for this great contribution.

Issue #610, opened 2018-11-06
grant_type  set to variable


mlboy:
“grant_type”  set to variable, because there are some non standard cases grant_type is not “authorization_code” some times it may be named "auth_code" ..

JonathanHuot:
Hi @mlboy, it seems a reasonnable feature.
However I'm wondering if it is not already doable.

Could you specify if you are talking about client, or server, and also an example of code and error ? 

It will be a lot more meaningful.

mlboy:
https://github.com/oauthlib/oauthlib/pull/611

mlboy:
such as Meitu account and some  non-standard internal  systems.

JonathanHuot:
Fixed in https://github.com/oauthlib/oauthlib/pull/625

PR #611, opened 2018-11-06
grant_type set to variable

mlboy:


JonathanHuot:
Is it possible to add the same kwargs processing for other flows ? e.g. `password`, `token` ?
Also, adding unit tests would be warmly appreciated.

skion:
When would someone use a grant type value different from the hard-coded one?

JonathanHuot:
@skion: @mlboy mentionned in https://github.com/oauthlib/oauthlib/issues/610, "Meitu" or internal systems were using custom grant names.

However, after more thinking, I am a bit dubious about the `**kwargs` choice. I'm not keen to change the signature of those methods (even if transparent in `**kwargs`) because it adds complexity where it should not. Also, the rare occurences of such probabilities is so low, that's not worth it.

Instead, to let be everyone happy, specially @mlboy :-), I'm suggesting you should rollback the `**kwargs` and implement it in a different way:
- change the hard-coded values (e.g.`"client_credentials"`) into a class variable (e.g. `BackendApplicationClient.client_credentials = "client_credentials"`) and use this variable instead of the hard-coded values.
- then, if anyone want to use a custom value against a custom-internal provider, it could easily change the class variable without changing the actual implementation, and examples, and unit tests, and so on and so forth...

E.g. 
```python
BackendApplicationClient.client_credentials = "cli_creds"  # was "client_credentials"

# do everything else, as usual
cli = BackendApplicationClient(xxx)
cli.prepare_request_body(..)
```

skion:
That definitely sounds like a better direction, thanks @JonathanHuot!

mlboy:
good idea, i will try it.  waiting me.

mlboy:
Is it better to use “grant_type”  or trait as attribute? 

BackendApplicationClient.grant_type  ?  BackendApplicationClient.client_credentials

just like this?

https://github.com/oauthlib/oauthlib/blob/fabcf865a7f33e5592453189670a82f08dc4384b/oauthlib/oauth2/rfc6749/clients/service_application.py#L32

JonathanHuot:
I think `grant_type` is better for the attribute name. Your example is correct, yes!

JonathanHuot:
Closed in favor of https://github.com/oauthlib/oauthlib/pull/625

PR #625, opened 2018-12-12
let grant_type support custom

mlboy:


mlboy:
#611

JonathanHuot:
@skion : It seems _this_ is applicable for OAuth2.0 Provider but not Client. The PR changed only the Client part. I guess that's enough?

Issue #650, opened 2019-02-13
Normalization of URIs for use in base strings does not percent encode spaces


hoylen:
**Describe the bug**

The normalisation of a URL does not process the space character (0x20) according to 
[section 3.4.1.2 of RFC 5849](https://tools.ietf.org/html/rfc5849#section-3.4.1.2).

The implementation is in the `normalize_base_string_uri` function in _oauthlib/oauth1/rfc5849/signature.py_.

**How to reproduce**

Invoke _normalize_base_string_uri_ with a URI that contains a space character (e.g. the example `http://EXAMPLE.COM:80/r v/X?id=123` from the RFC, there is a space between the "r" and the "v").

The returned value is `http://example.com/r v/X`, where the space is not percent encoded.

**Expected behavior**

The returned value should have the space percent encoded: `http://example.com/r%20v/X`

**Additional context**

Please provide any further context here.

- Are you using OAuth1, OAuth2 or OIDC?

This was discovered when using OAuth1 with RSA-SHA1. The signature verification was failing, because the base string that was signed is different between the client and the server. The client was signing a base string that had the space percent encoded, but the server used a base string that didn't have the space encoded correctly. Note: the correct base string encodes the normalised URI, so it actually will contain "%2520" (the percent encoding of "%20").

- Are you writing client or server side code?

Server side code.

Python unit test code, using examples from the RFC:
- [test_rfc5849.py](https://github.com/oauthlib/oauthlib/files/2858847/test_rfc5849.txt)

JonathanHuot:
Hi @hoylen, are you able to make a PR for this ? It will be very helpful as you already made a good investigation around.

PR #651, opened 2019-02-19
Fixed space encoding in base string URI used in the signature base st…

hoylen:
Encode spaces in the URI correctly as "%20" in the _base string URI_.
This pull request addresses issue <https://github.com/oauthlib/oauthlib/issues/650>.

The _base string URI_ is used in the _signature base string_, which is the value that is signed. Before this fix, any spaces were not encoded and that caused the wrong signature to be generated, or for signature verification to fail.

Comments have been added to point out that RFC 5849 is deficient in this area, and the only indication that spaces must be encoded is in one of its examples. It is not known whether other special characters need to be encoded too.

The function "normalize_base_string_uri" has been renamed to `base_string_uri`, since it just calculates the [base string URI](https://tools.ietf.org/html/rfc5849#section-3.4.1.2). There is no "normalisation" (whatever that means) of the "base string URI" value. Suggestion: rename _construct_base_string_ to _construct_signature_base_string_ or _signature_base_string_ (since it calculates the [signature base string](https://tools.ietf.org/html/rfc5849#section-3.4.1), not something called a "base string").

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+TBD
+---
+
+* Fixed space encoding in base string URI used in the signature base string.
+
 3.0.0 (2019-01-01)
 ------------------
 OAuth2.0 Provider - outstanding Features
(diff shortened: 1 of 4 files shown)


JonathanHuot:
Hi @hoylen, could you see why tests are failing ? You can run `tox -e <yourpyversion>` on your environment to debug. Thanks !

hoylen:
I missed one change. Line 136 in _oauthlib/oauth1/rfc5849/__init__.py_ needs "normalize_base_string_uri" changed to "base_string_uri". Now fixed.

Issue #652, opened 2019-02-19
OIDC id_token response should not contain `state`


JonathanHuot:
**Describe the bug**

When requesting an id_token with the authorization code flow, the TOKEN endpoint returns a `state`, but `state` is only a response's parameter for the AUTHORIZE endpoint.

**How to reproduce**

- Implement OIDC support in oauthlib for `authorization_code`.
- Send a valid `/authorize` request with a state
- Send a valid `/token` request, and see `state=None` in the response.

**Expected behavior**

We should have the state only in the `/authorize` response (e.g. for code or implicit's response).

**Additional context**
- Are you using OAuth1, OAuth2 or OIDC?
OIDC

- Are you writing client or server side code?
Server side

JonathanHuot:
The same bug exists for OAuth2 but is slightly different. The response of `grant_type=authorization_code&state=foobar` contains `state=foobar`. However, we don't have the field if the request doesn't contain it, it's why nobody already noticed it.

It is like this because the same code is run for implicit and for authorization_code. This should be fixed.

PR #653, opened 2019-02-20
Fix 652: removed "state" from /token response.

JonathanHuot:
Fix OIDC `/token` flow where `&state=None` was always returned, and fix OAuth2.0 `/token` flow where `&state=foobar` was returned if `&state=foobar` was present in the token request.

Removed `save_token` argument from `create_token()` signature cuz it was not used internally. Deprecated the option to let upstream libraries have a chance to remove it, if ever used.

### DIFF
--- a/tests/oauth2/rfc6749/endpoints/test_credentials_preservation.py
+++ b/tests/oauth2/rfc6749/endpoints/test_credentials_preservation.py
@@ -42,18 +42,6 @@ class PreservationTest(TestCase):
 
     def test_state_preservation(self):
         auth_uri = 'http://example.com/path?state=xyz&client_id=abc&response_type='
-        token_uri = 'http://example.com/path'
-
-        # authorization grant
-        h, _, s = self.web.create_authorization_response(
-                auth_uri + 'code', scopes=['random'])
-        self.assertEqual(s, 302)
-        self.assertIn('Location', h)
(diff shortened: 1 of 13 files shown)


JonathanHuot:
Fix https://github.com/oauthlib/oauthlib/issues/652

tests/oauth2/rfc6749/endpoints/test_credentials_preservation.py
```diff
@@ -42,18 +42,6 @@ def set_client(self, request):
...
     def test_state_preservation(self):
         auth_uri = 'http://example.com/path?state=xyz&client_id=abc&response_type='
-        token_uri = 'http://example.com/path'
-
-        # authorization grant
-        h, _, s = self.web.create_authorization_response(
-                auth_uri + 'code', scopes=['random'])
-        self.assertEqual(s, 302)
-        self.assertIn('Location', h)
-        code = get_query_credentials(h['Location'])['code'][0]
```
skion: This (the authorization response) redirect, *should* contain a `state` as a query parameter if the authorization request included one. But I guess that is tested elsewhere already?

JonathanHuot: No I had a bit overlooked and I removed too many lines :-) Let me fix thiw test there.

PR #658, opened 2019-02-28
Removed duplicated OIDC members in OAuth2.RequestValidator

JonathanHuot:
All `openid.RequestValidator` methods were duplicated in `oauth2.RequestValidator`, which is not natural when implementing a plain "OAuth2.0" RequestValidator.

Also, it is more consistent when migrating the oauthlib implementation from OAuth2 to OpenID Connect, we have to switch to the OIDC `Server` but also to the OIDC `RequestValidator`, meaning implementing the new OIDC methods.

### DIFF
--- a/oauthlib/oauth2/rfc6749/request_validator.py
+++ b/oauthlib/oauth2/rfc6749/request_validator.py
@@ -291,32 +291,6 @@ class RequestValidator(object):
         """
         raise NotImplementedError('Subclasses must implement this method.')
 
-    def get_authorization_code_scopes(self, client_id, code, redirect_uri, request):
-        """ Extracts scopes from saved authorization code.
-
-        The scopes returned by this method is used to route token requests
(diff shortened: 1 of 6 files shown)


oauthlib/oauth2/rfc6749/request_validator.py
```diff
@@ -378,104 +352,6 @@ def save_bearer_token(self, token, request, *args, **kwargs):
         """
         raise NotImplementedError('Subclasses must implement this method.')
 
-    def get_jwt_bearer_token(self, token, token_handler, request):
```
skion: Could this one not be used in an OAuth 2.0 flow?

JonathanHuot: Indeed, however I'd suggest to change it again the other way around if ever it's added into an OAuth2 grant (currently it is only called from openid grant).

PR #663, opened 2019-03-21
fix include_client_id argument

akeeman:
None values are passed to `oauthlib.oauth2.rfc6749.parameters.prepare_token_request`'s `include_client_id` argument, while this function only accepts booleans. Taken documentation and default value from `prepare_token_request` and updated the clients.
Problems are caused by b4ceb8a7

An alternative is to make None a valid input again, and include some auto behaviour on whether to include client_id or not.

### DIFF
--- a/oauthlib/oauth2/rfc6749/clients/backend_application.py
+++ b/oauthlib/oauth2/rfc6749/clients/backend_application.py
@@ -29,11 +29,11 @@ class BackendApplicationClient(Client):
     Since the client authentication is used as the authorization grant,
     no additional authorization request is needed.
     """
-    
+
     grant_type = 'client_credentials'
-    
+
     def prepare_request_body(self, body='', scope=None,
-                             include_client_id=None, **kwargs):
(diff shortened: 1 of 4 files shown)


oauthlib/oauth2/rfc6749/clients/backend_application.py
```diff
@@ -29,11 +29,11 @@ class BackendApplicationClient(Client):
...
     no additional authorization request is needed.
     """
-    
+
     grant_type = 'client_credentials'
-    
+
     def prepare_request_body(self, body='', scope=None,
-                             include_client_id=None, **kwargs):
+                             include_client_id=True, **kwargs):
```
JonathanHuot: Shouldn't be default to `False` to be backward compatible ?

thedrow: I agree.

akeeman: Yes, it should.

Issue #666, opened 2019-04-19
Client_secret and code_verifier (PKCE) should be transmitted securely


polamayster:
**`client_secret` and `code_verifier` are accepted when sent as parameters in query string**

`Request.client_secret` should be checked for presence in headers or body and `Request.code_verifier` just in body but not query string as it is sensitive data.
Addition checks might be done, such as request type is `POST` and data was sent using `HTTPS`.

When `client_secret` or `code_verifier` is sent in query string it should result in Bad Request, enforcing client to send data securely.

JonathanHuot:
Hi @polamayster, you're perfectly right.

I'm adding the sections of RFC specifying how it has to be implemented:

Section of OAuth2.0 RFC :

https://tools.ietf.org/html/rfc6749#section-2.3.1

```
2.3.1.  Client Password
   Clients in possession of a client password
   MAY use the HTTP Basic authentication scheme (..) 

   Alternatively, the authorization server
   MAY support including the client credentials in the request-body (..)

   The parameters can only be transmitted in the request-body and
   MUST NOT be included in the request URI.
```

Section of PKCE RFC:

https://tools.ietf.org/html/rfc7636#section-4.5

```
4.5.  Client Sends the Authorization Code and the Code Verifier to the
      Token Endpoint
    In addition to the parameters defined in the OAuth 2.0 Access
    Token Request (Section 4.1.3 of [RFC6749]), it sends the following parameter:

   code_verifier
      REQUIRED.  Code verifier
```

https://tools.ietf.org/html/rfc6749#section-4.1.3
```

4.1.3.  Access Token Request

   The client makes a request to the token endpoint by sending the
   following parameters (..) in the HTTP request entity-body:
```

So I have no doubt about the pertinence of the issue. Any PRs are welcome!

Abhishek8394:
I'll be interested in taking this on.

Abhishek8394:
Should this behaviour be enforced only for requests that expect credentials or any request in general?

polamayster:
To my understanding `oauthlib.common.Request` class has a `__getattr__` method and `_params` dictionary (updated from query string and body parameters) and **any request** that tries to access/get `client_secret` or `code_verifier` should only lookup in body and/or headers (perhaps having separate property for sensitive data lookup would make sense)

Abhishek8394:
I see, I'll submit a pr for it today

Abhishek8394:
[...]

PR #667, opened 2019-04-21
Fix Issue #666: ban 'client_secret' and 'code_verifier' from url query params

Abhishek8394:
Fixes for issue #666 
- Checks that ~'client_secret' and 'code_verifier' do not appear as url query params~  there are no query params
- ~Checks are made irrespective of what kind of request it is and if those params will ever be accessed or not.~  Enforce POST HTTP method on TokenEndpoint, IntrospectEndpoint and RevocationEndpoint


Abhishek8394:
I agree about keeping it out of Request class, I was simply following the comments from the related issue. Although regrading the below paragraph:

> As an example, if we use a `ResourceEndpoint/verify_request` with `client_secret` in its Query URI, the `oauthlib.Request` object will raise a `ValueError` but using `client_secret` for a resource access can be perfectly valid for a given webserver !

If clients want to access a resource with their credentials, the recommended way is to send HTTP basic auth header; or get an access token for the resource with client credentials and use that to access the resource. And I think this approach applies to all requests but there are might be edge cases like quick debugging where secret in query parameters is desired. In that case, we either enforce rules on selected protected endpoints or leave it as is.

Abhishek8394:
Quick question, I have the checks activate only for POST requests on specified endpoint. This was an attempt at being explicit about the conditions. Should we cover other HTTP methods on those endpoints or not?

JonathanHuot:
Hi @Abhishek8394, thanks for continuing on this, that's very helpful.

Only POST are valid for these three endpoints. I think we can even raise an `invalid_request` error if we receive a GET/PUT/...

Abhishek8394:
> Hi @Abhishek8394, thanks for continuing on this, that's very helpful.
> 
> Only POST are valid for these three endpoints. I think we can even raise an `invalid_request` error if we receive a GET/PUT/...

About this, I have about 60 tests failing with this change. I'm not sure if it's just tests or do actually lot of existing use cases rely on this incorrect behaviour. My guess is the main culprit has to be `GET` requests being used on these endpoints, which will break anyway with our ban on query parameters. Let me know if you still wish to proceed with that change. I can link you to a branch of those changes or email a patch if you're interested in checking it out yourself.

JonathanHuot:
Hi @Abhishek8394 , could you join our chat https://gitter.im/oauthlib/Lobby to discuss 'bout that ? and yes, I'm curious to see those & see if we can avoid that :-) (however if we can't avoid it and it looks going in the right direction, let's do it)

Abhishek8394:
So the large number of tests failing was due to a bug introduced by me (oops). Anyway, now all tests pass again, no existing tests needed to be altered.

oauthlib/oauth2/rfc6749/endpoints/base.py
```diff
@@ -62,6 +66,16 @@ def _raise_on_unsupported_token(self, request):
             request.token_type_hint not in self.supported_token_types):
             raise UnsupportedTokenTypeError(request=request)
 
+    def _raise_on_bad_post_request(self, request):
+        """Raise if invalid POST request received
+        """
+        if request.http_method.lower() == 'post':
+            query_params = CaseInsensitiveDict(dict(urldecode(request.uri_query)))
+            for param in self._blacklist_query_params:
```
JonathanHuot: Maybe a last comment for me: we agreed that the specifications require to have all OAuth2 POST parameters in the body and not in the query.

So, instead of focusing only on the "sensitive" parameters like secret+verifier, what about if you replace `self._blacklist_query_params` by `request._params.keys()` ? 

That way, we are again closer to the specification. 
Thoughts?

Abhishek8394: I just pushed with banning any and all query parameters; on the target endpoints only of course, not just OAuth related. Would that work?

Abhishek8394:
And now bandit is going nuts over method parameter defaults that I did not touch. Should I try to suppress those with `#nosec`? I've never dealt with this before.

JonathanHuot:
I had the same issue on another PR, let's wait for mine to be merged and it
should be ok (or cherry-pick the commit on the other PR?)

Le mar. 14 mai 2019 10:07, Abhishek Patel <notifications@github.com> a
écrit :

JonathanHuot:
Don't forget https://github.com/oauthlib/oauthlib/pull/677/commits/f25b9133646d0120f6221e9d7cbf16e747dd6aa7 also :-)

oauthlib/oauth2/rfc6749/endpoints/base.py
JonathanHuot: That's great!

mastizada:
Can anyone please explain why all query parameters were banned in POST requests? What was the idea behind that? We use a site request parameter for identifying the source and after the version update, the api is broken and now we have to fork the repository :(

This parameter is parsed directly in the middleware and is not checked on views, so it doesn't depend on the method type. We can adjust the check for the post data, but we can't ask all our clients to update their usage in a short time.

thedrow:
I'm assuming we didn't want to explicitly exclude each and every dangerous query parameter.

If you need to fork to support your use-case then we have a design issue and you should open an issue about it.

mastizada:
@thedrow thanks for info, how I can rewrite `_raise_on_bad_post_request`? I use a django plugin that is using this library. If it is not possible to rewrite this way, I already added whitelist support using environment variables, can open PR with it.

thedrow:
Which Django plugin? I can help but you need to provide me with some details.
Come to our Gitter channel. I promise I'll do whatever is needed to help you :)

mvaled:
@mastizada @thedrow 

We've recently come to the same issue.  We're upgrading an old application to Django 3.1 (with django-oauth-toolkit>=1.3.3 which requires oauthlib>=3.1).

We need to keep the old-way of doing things for a while to allow clients of the application to update without breaking the API.  Have you come to a solution?

thedrow:
Sorry I no longer recall.

Issue #665, opened 2019-04-05
Security data leak to logs


lietu:
**Describe the bug**

OAuthLib leaks users' passwords etc. to logs in case of errors.

We have e.g. an "authorize" route, that takes "username" + "password" in JSON body. However, if there's errors in parsing the request or validating some information, their username + password get leaked into the error logs.

OAuthLib should not make ANY assumptions about my request structure and it should NEVER leak the complete request body as part of an exception message or otherwise.

Example of leak when using a `bottle` application with `bottle-oauthlib`, exception generated by `oauthlib` contains request body and so the username and password when an error is caused by redirect URI mismatch.

```
172.17.0.9 - - [05/Apr/2019 08:16:03] "POST /authorize?response_type=code&client_id=2fcc9e52-b29e-412c-ad5b-a74734a35fc7&redirect_uri=http%3A%2F%2Fworld.local%3A32000%2Fapi%2FexchangeToken_&grant_type=authorization_code&state=123qwerasdf HTTP/1.0" 500 62
Traceback (most recent call last):
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/bottle.py", line 862, in _handle
    return route.call(**args)
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/bottle.py", line 1740, in wrapper
    rv = callback(*a, **ka)
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/webargs/core.py", line 450, in wrapper
    return func(*new_args, **kwargs)
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/bottle_oauthlib/oauth2.py", line 237, in wrapper
    raise e
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/bottle_oauthlib/oauth2.py", line 230, in wrapper
    uri, http_method=http_method, body=body, headers=headers, scopes=scope
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/oauthlib/oauth2/rfc6749/endpoints/base.py", line 64, in wrapper
    return f(endpoint, uri, *args, **kwargs)
  File "/.venv/auth-ALEBwC_u/lib/python3.6/site-packages/oauthlib/oauth2/rfc6749/endpoints/authorization.py", line 104, in create_authorization_response
    request, self.default_token_type)
[...]

PR #668, opened 2019-04-21
Fix issue #665: Add method to get / set debug flag (#668)Merge pull request #668 from Abhishek8394/issue-665

Abhishek8394:
Fixes issue #665 
- By default debug mode is always off
- Debug mode turned on automatically for tests
- Request `repr` output completely sanitized in non debug mode
- Modules can check for debug mode with `oauthlib.get_debug_flag` method.

### DIFF
--- a/oauthlib/__init__.py
+++ b/oauthlib/__init__.py
@@ -19,8 +19,16 @@ logging.getLogger('oauthlib').addHandler(NullHandler())
 _DEBUG = False
 
 def set_debug(debug_val):
+	"""Set value of debug flag
+	
+    :param debug_val: Value to set. Must be a bool value.
+	"""
 	global _DEBUG
 	_DEBUG = debug_val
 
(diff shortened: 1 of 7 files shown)


oauthlib/__init__.py
```diff
@@ -15,3 +15,20 @@
...
+
+def set_debug(debug_val):
+	"""Set value of debug flag
+	
+    :param debug_val: Value to set. Must be a bool value.
+	"""
+	global _DEBUG
+	_DEBUG = debug_val
+
+def get_debug_flag():
```
JonathanHuot: `set_debug` & `get_debug_flag` should have symmetric names. I'd suggest `set_debug` / `get_debug` ? other suggestions are ok, but should be symmetric.

Abhishek8394: Refactored it.

PR #674, opened 2019-04-30
token_type should be case insensitive

bungoume:
According to RFC, token_type is case insensitive.
https://tools.ietf.org/html/rfc6749#section-4.2.2

Some softwares check Bearer as case insensitive.
https://github.com/keycloak/keycloak/blob/b478472b3578b8980d7b5f1642e91e75d1e78d16/services/src/main/java/org/keycloak/services/managers/AppAuthManager.java#L50

This pull request allows case insensitive `bearer` string.

### DIFF
--- a/oauthlib/oauth2/rfc6749/tokens.py
+++ b/oauthlib/oauth2/rfc6749/tokens.py
@@ -254,7 +254,7 @@ def get_token_from_header(request):
 
     if 'Authorization' in request.headers:
         split_header = request.headers.get('Authorization').split()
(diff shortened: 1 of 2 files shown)


JonathanHuot:
Hi @bungoume, it looks a simple and useful one. Do you mind to simple unit test to confirm it is working as expected ?
That would be great.

bungoume:
Thank you for confirmation.

I restored the test case (added at #491 and deleted in https://github.com/oauthlib/oauthlib/pull/525/commits/07aef4ed317d473c3e86d58ace9d9813a37de1dc),
and added simple lowercase test.

PR #675, opened 2019-04-30
Handle null value in expires_in field in JSON handler

### LINKED ISSUE #672: Null value in expires_in field not handled correctly
**Describe the bug**

Some OAuth-based services have been discovered to include the `expires_in` field, but set it to `null`. This issue was discovered on our end attempting to authenticate to SurveyMonkey's API. The `oauthlib` library does not handle this case correctly, and will throw an error if it encounters this.

**How to reproduce**

1. Create an account with SurveyMonkey
1. Create a sample app to test against
1. Attempt to authenticate using OAuth flow against the sample app
1. Receive the following error: `int() argument must be a string, a bytes-like object or a number, not 'NoneType'`

**Expected behavior**

Ignore the null `expires_in` field, and continue without throwing an error.

**Additional context**

The following is an example of a returned payload (access_token hidden for security purposes):

`{"access_token":".g.w2***kjP","token_type":"bearer","access_url":"https:\/\/api.surveymonkey.com","expires_in":null}`

The following is the stacktrace we receive:

```
  File "/code/ac_auth/views.py", line 59, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 495, in dispatch
    response = self.handle_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 455, in handle_exception
    self.raise_uncaught_exception(exc)
  File "/usr/local/lib/python3.7/site-packages/rest_framework/views.py", line 492, in dispatch
    response = handler(request, *args, **kwargs)
  File "/code/ac_auth/views.py", line 158, in get
    request.get_raw_uri(),
  File "/code/ac_auth/oauth_service.py", line 46, in fetch_access_token
    timeout=CONNECTION_TIMEOUT,
  File "/usr/local/lib/python3.7/site-packages/requests_oauthlib/oauth2_session.py", line 307, in fetch_token
    self._client.parse_request_body_response(r.text, scope=self.scope)
  File "/usr/local/lib/python3.7/site-packages/oauthlib/oauth2/rfc6749/clients/base.py", line 415, in parse_request_body_response
    self.token = parse_token_response(body, scope=scope)
  File "/usr/local/lib/python3.7/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 422, in parse_token_response
    params['expires_at'] = time.time() + int(params['expires_in'])
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```

We are using `oauthlib` 3.0.1 from pip.

- Are you using OAuth1, OAuth2 or OIDC? **OAuth2**
- Are you writing client or server side code? **Server-side**
- If client, what provider are you connecting to? **N/A**
- Are you using a downstream library, such as `requests-oauthlib`, `django-oauth-toolkit`, ...? **requests-oauthlib**

shssoichiro:
Closes #672


shssoichiro:
I'm not 100% is this is the ideal way we'd like to handle this edge case. I decided to remove `expires_in` if it's `null` because that provides consistency to any functions that rely on the parsed response.

JonathanHuot:
Hi @shssoichiro, RFC says that if `expires_in` is not present, the Authorization Server should use a default value of expiration (e.g. mentionned in the docs). So, if we remove `expires_in` when `expires_in is None`, it basically means that `expires_in=None` is the same as `expires_in=<default_value>`. 
However, I think that `expires_in=None` _could_ also says _never_ expires for instance.

I think because of the uncertainty of the `None` value, I think we have to keep it present and not remove the field.

Others ?

shssoichiro:
@JonathanHuot I'm open to making that change, depending on how we want to represent it. You asked for input from others on this project, and I'd like to see what the general consensus is as well.

thedrow:
Are tokens allowed not to expire?
I don't think so since this may pose a permanent security risk.

thedrow:
```
expires_in
         RECOMMENDED.  The lifetime in seconds of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value.
```

So if it's not provided in any way, we should resort to a known default.
LGTM.

shssoichiro:
The reason I chose this approach as well as that I suspected SurveyMonkey's `null` response may be due to a difference in how their REST framework serializes nulls. I've seen that some REST frameworks will serialize `null` fields to JSON as a JSON `null` literal, whereas other REST frameworks will serialize it by excluding the field from the JSON output. But given this, if there is no explicit meaning to a `null` expires_in in the spec, which it does not seem that there is, I think it makes the most sense to treat `null` and an absent value as the same.

PR #680, opened 2019-06-06
Check for errors in authorization code response

### LINKED ISSUE #290: `parse_authorization_code_response` has no error checking
Over at https://github.com/gratipay/gratipay.com/issues/2870 we're finding that [`parse_authorization_code_response`](https://github.com/idan/oauthlib/blob/04631866c058defb480462db1ec13cb7f88780a9/oauthlib/oauth2/rfc6749/parameters.py#L179-L229) sometimes receives an error response that it's [not prepared to handle](https://github.com/idan/oauthlib/blob/04631866c058defb480462db1ec13cb7f88780a9/oauthlib/oauth2/rfc6749/parameters.py#L221-L223). Here's a couple examples of the querystrings we're seeing from different providers:

```
Google:   error=access_denied&state=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
Facebook: error=access_denied&error_code=200&error_description=Permissions+error&error_reason=user_denied&state=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

Should a check for `error` be added, with a call to [`raise_from_error`](https://github.com/idan/oauthlib/blob/04631866c058defb480462db1ec13cb7f88780a9/oauthlib/oauth2/rfc6749/errors.py#L249-L259) (but cf. #2869)?

MrkGrgsn:
Fixes #290 by adding error checking to `parse_authorization_code_response()` before any validation of the `code` param is done.

Note that I also shifted the `state` check before both the error and code checks - `state` is required in the response and if it isn't correct, why would we trust any of the other params?  This is a little out-of-scope so let me know if you'd prefer I revert that bit.

Also note that this is a breaking change as I mentioned in https://github.com/oauthlib/oauthlib/issues/290#issuecomment-499331868 because client error handling behaviour will need to change.

### DIFF
--- a/oauthlib/oauth2/rfc6749/parameters.py
+++ b/oauthlib/oauth2/rfc6749/parameters.py
@@ -264,11 +264,14 @@ def parse_authorization_code_response(uri, state=None):
     query = urlparse.urlparse(uri).query
     params = dict(urlparse.parse_qsl(query))
 
-    if not 'code' in params:
-        raise MissingCodeError("Missing code parameter in response.")
-
     if state and params.get('state', None) != state:
         raise MismatchingStateError()
+
+    if 'error' in params:
+        raise_from_error(params.get('error'), params)
+
+    if not 'code' in params:
+        raise MissingCodeError("Missing code parameter in response.")
 
     return params
 
--- a/tests/oauth2/rfc6749/test_parameters.py
+++ b/tests/oauth2/rfc6749/test_parameters.py
@@ -73,7 +73,8 @@ class ParameterTests(TestCase):
     error_nocode = 'https://client.example.com/cb?state=xyz'
     error_nostate = 'https://client.example.com/cb?code=SplxlOBeZQQYbYS6WxSbIA'
     error_wrongstate = 'https://client.example.com/cb?code=SplxlOBeZQQYbYS6WxSbIA&state=abc'
-    error_response = 'https://client.example.com/cb?error=access_denied&state=xyz'
+    error_denied = 'https://client.example.com/cb?error=access_denied&state=xyz'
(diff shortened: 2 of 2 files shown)


JonathanHuot:
Hi @MrkGrgsn !

This looks a good change to me.

However, before approving it, and in order to add a little bit of context, could you tell which library are you using and/or how do you use this library ? (sample of code)

Thanks !

MrkGrgsn:
Hi @JonathanHuot 

I'm working on an [ORCID](https://orcid.org/) integration with a Django app using requests_oauthlib. The goal is to link the local user account with their ORCID record and later query and retrieve some resources. I'm using the 'Web Application Flow' aka Authorisation Code Grant Type. 

I can't link to the source code but the code pre-PR is similar to this simplified example showing the view that handles the redirect from the authorisation server:

```python
class FetchAccessTokenView():
    def get():
        try:
            token = session.fetch_token()
        except MissingCodeError:
            error = request.GET.get('error', None)
            if error:
                if error == 'access_denied':
                    # ask user to reconsider
                elif error == 'temporarily_unavailable':
                    # ask user to try again later
            else:
                #  unexpected, log and provide feedback
```
Which will break post-PR.  On the other hand, you could check the error param prior to calling `fetch_token` and this would not break post-PR but we're still doing work that oauthlib can do.

With the attached PR, I can simply use the try/except without any extra handling of the error param:

```python
class FetchAccessTokenView():
    def get():
        try:
            token = session.fetch_token()
        except AccessDeniedError:
            # ask user to reconsider
        except TemporarilyUnavailableError
            # ask user to try again later
        except MissingCodeError:
            # unexpected, log and provide feedback
```

Issue #603, opened 2018-10-18
Improve BackendApplicationClient class


DiegoQueiroz:
**Describe the feature**

In `BackendApplicationClient` class, the `client_id`, `scope`, and `**kwargs` parameters passed to the constructor are useless. Notice that even `client_id` being mandatory, it is not used at all (it is only stored). The `**kwargs` parameter, on the other hand, is completely ignored.

The `prepare_request_body` method overrides `scope` and ignores the one passed to the constructor. The same happens to `**kwargs`. This forces me to pass `client_id`/`scope`/`**kwargs` twice (actually, only once, since I can pass anything to the constructor, the result will be the same).

**Additional context**

My suggestion is to override the constructor in the `BackendApplicationClient` subclass with this code (or something similar):

```
from __future__ import absolute_import, unicode_literals

from ..parameters import parse_token_response, prepare_token_request
from oauthlib.common import generate_token
from .base import Client, AUTH_HEADER

class BackendApplicationClient(Client):
    def __init__(self, client_id,
                 default_token_placement=AUTH_HEADER,
                 token_type='Bearer',
                 access_token=None,
                 refresh_token=None,
                 mac_key=None,
                 mac_algorithm=None,
                 token=None,
                 scope=None,
                 state=None,
                 redirect_url=None,
                 state_generator=generate_token,
                 **kwargs):
        # forward parameters to super class
        super(BackendApplicationClient, self).__init__(client_id,
                 default_token_placement,
                 token_type,
                 access_token,
                 refresh_token,
                 mac_key,
                 mac_algorithm,
                 token,
                 scope,
                 state,
                 redirect_url,
                 state_generator,
                 **kwargs)
        # store kwargs (client_id and scope are already stored by super class method)
        # better still would be to add this piece of code directly in the `Client` superclass
        self.kwargs = kwargs

    def prepare_request_body(self, body='', scope=None, **kwargs):
        if scope is None:
            # no new scope provided, use the class scope
            scope = self.scope
        if not kwargs:
            # empty kwargs, use class kwargs
            kwargs = self.kwargs
        if self.client_id and 'client_id' not in kwargs:
            # kwargs does not contains client_id but the class has one, add it
            kwargs['client_id'] = self.client_id
        return prepare_token_request('client_credentials', body=body,
                                     scope=scope, **kwargs)
```

This way, the parameters passed to the constructor are used for something, while we keep compatibility with existing code.

JonathanHuot:
[...]

PR #682, opened 2019-07-02
Fix BackendApplicationClient.prepare_request_body

qporest:
Currently, if no `scope` is passed to `prepare_request_body`, None will be passed on to `prepare_token_request`, even if BackendApplicationClient was initialized with `scope`.


JonathanHuot:
LGTM, related to https://github.com/oauthlib/oauthlib/issues/603

Issue #695, opened 2019-08-10
OAuth1 Provider Question on urlencoding oauth_callback before checking signature


ashleysommer:
I'm implementing a proof-of-concept OAuth 1.0 provider and client, and I'm using oauthlib to help write the provider side.

I'm having the problem in signature matching, where I cannot produce identical basestrings in the client vs in the provider.

In the client, the generated basestring looks like this:
```
GET&http%3A%2F%2Flocalhost%3A8098%2Foauth%2Frequest_token&oauth_callback%3Dhttp%3A%2F%2F127.0.0.1%3A8888%2Foauth%26oauth_consumer_key%3Ddev%26oauth_nonce%3D5661035ff068a8c93465a8bbf96b8d%26oauth_signature_method%3DHMAC-SHA1%26oauth_timestamp%3D1565436342%26oauth_version%3D1.0%26realm%3Demail`
```
Whereas in the provider, the generated basestring looks like this:
```
GET&http%3A%2F%2Flocalhost%3A8098%2Foauth%2Frequest_token&oauth_callback%3Dhttp%253A%252F%252F127.0.0.1%253A8888%252Foauth%26oauth_consumer_key%3Ddev%26oauth_nonce%3D5661035ff068a8c93465a8bbf96b8d%26oauth_signature_method%3DHMAC-SHA1%26oauth_timestamp%3D1565436342%26oauth_version%3D1.0%26realm%3Demail`
```
As you can see, they are _almost_ identical, except the oauth_callback parameter:
oauth_callback=`http%3A%2F%2F127.0.0.1%2Foauth`
vs
oauth_callback=`http%253A%252F%252F127.0.0.1%252Foauth`
It looks like oauthlib double-urlencodes it. (`%25` is urlencoded form of `%` character)

You can see the query key-value pairs get urlencoded first during the normalisation process here: https://github.com/oauthlib/oauthlib/blob/31486aa503e19ae26ba4c15f2809ddf6c5c1b248/oauthlib/oauth1/rfc5849/signature.py#L423

Then you can see the whole normalised querystring is urlencoded _again_ when creating the basestring here: https://github.com/oauthlib/oauthlib/blob/31486aa503e19ae26ba4c15f2809ddf6c5c1b248/oauthlib/oauth1/rfc5849/signature.py#L104

The client code only urlencodes the whole query string once before generating the basestring.

[...]

PR #697, opened 2019-08-14
Fix tests for OAuth1 signature basestring generation,

ashleysommer:
Fix tests for OAuth1 signature basestring generation to better align with examples and expected results set out in the RFC doc.

Fixes https://github.com/oauthlib/oauthlib/issues/695


thedrow:
Your changes cause a failure on Python 2.7.
It could be a bytes/unicode issue.

Can you please take a look?

ashleysommer:
Yes, its probably a bytes/unicode thing. I will look into when I get a chance. Maybe tomorrow.

ashleysommer:
Fixed the failing test on python 2.7

PR #700, opened 2019-08-15
Drop support for Python 2.7

### LINKED ISSUE #698: Drop Python 2.7
Python 2.7's EOL is in January 2020.
We should be prepared to drop the support for it.

This will simplify our codebase and allow us to use newer features.

hugovk:
Fixes #698.


ashleysommer:
Fixed the failing test on python 2.7

thedrow:
@JonathanHuot Do we want this in 4.0?

JonathanHuot:
I'm a bit worried about the downstream libraries e.g. requests-oauthlib, pythohn-social-auth, many others which are still testing with py2.7

thedrow:
They'll have to drop it soon anyway.

Issue #686, opened 2019-07-16
Support for Azure AD on-behalf-of grant


steffenschumacher:
**Describe the feature**
Support for https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow even though it its not a part of the standard?

If we add an OnBehalfOfApplicationClient derived from Client, it would easily support the token grant flow described above - there'd be very few changes compared to eg WebApplicationClient

**Additional context**
I have implemented my own client, and are happy, but if there is no issue with supporting proprietary grants, it could be made available for all?

This client-side feature would apply both to OAuth1 and 2, although I've only used it myself with 2.

JonathanHuot:
I guess it can be integrated, I think it is a good idea, even if not standard, blabla, we may have other interested.

thedrow:
@JonathanHuot If it's not a standard, maybe we can offer it in a third-party?

JonathanHuot:
Hi @thedrow, I'm not much familiar with the requests-oauthlib/compliances folder, but could it be there ?

Also, after digging into the docs, it seems that Azure describes it as an OAuth2 flow, not OAuth1.

thedrow:
It's your call.
In my opinion we should only include something here if it is commonly used.
Otherwise, it should be in a different package.

PR #640, opened 2018-12-28
TODO: Provider - Device code

jcampbell05:
This PR focuses on the server implementation of #626 and leaves the client implementation for a future PR. Since its in a draft I have placed it in a draft folder to indicate this so those who need it can use it with the awareness it may change in the future and will be re-named when the specification is stable.

Device Code is a new authorization flow which allows a client to be authenticated by a user for a device with limited or no input. 

I based this implementation on the authorization grant type since this flow shares some similarities but I utilized some concepts for the client grant type where this flow differs.

To understand how to implement the new classes and flow for a server here is an overflow of the flow and how a server would use them:

1. The server implements an authorization endpoint the client can use to request a user & device code. When this endpoint is hit the DeviceCodeGrant's `create_authorization_response` is called. A DeviceToken should be passed in with the expiry date configured by the server as well as a RequestValidator configured to return the verification url and polling interval. 
2. The client displays the user code to the user and the verification url. Once user enters code which exists . The server calls DeviceCodeGrant's `create_verification_response` which uses the validator to check that code exists.
3. If it exists then the server will ask user to allow or disallow. The server calls DeviceCodeGrant's `create_verification_response` which will this time ask the validator to update the authorized state.
4. Whilst this happening the client is polling a endpoint for the token status. Each time the server will call DeviceCodeGrant's `create_token_response` method. The validator's validate_token_request will validate the device token is authorized and should implement all of the error codes from the RFC (including rate limiting if needed).

Once authorized this method will use the create_token method on the BearerToken to create a new Access Token, ask the validator to save it and then invalidate the old device code.

Issue #111, opened 2013-02-13
OAuth 2 Server - RFC7523 - Assertions


ib-lundgren:
As per http://tools.ietf.org/html/draft-ietf-oauth-assertions-10

A new and from a brief look a very generic grant type with two orthogonal purposes, client auth and as auth grants. Need to check spec maturity and research whether it can simply be added as 1 or 2 new grant types.

JonathanHuot:
The RFC7523 specification is final since a couple of years now, and the overall spec is split into the Client Authentication part and the Authorization Grant part. Also Client vs Provider distinction should be done.

Authorization Grant part is implemented on client side here: https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/clients/service_application.py and we have a current ticket worth to check if this issue progress: https://github.com/oauthlib/oauthlib/issues/502

The Client Authentication part allows to replace client_id+client_secret authentication with an JWT assertion. In term of OAuth2.0 "client" metadata, it means that it adds new values to the `token_endpoint_auth_method`. Existing are: 

- none
- client_secret_post
- client_secret_basic

New from OIDC spec are:
- client_secret_jwt
- private_key_jwt


NB: Note that RFC7522 exists for SAML2 assertions, so the code must be easily extendable to this and later extensions thru the generic RFC7521.

PR #702, opened 2019-10-18
Initial custom grant type documentation.

JonathanHuot:
Improved Grant Type section to let developers create or implement their own custom grant type. Or also help them implementing new RFC.

Related to https://github.com/oauthlib/oauthlib/issues/111 and https://github.com/oauthlib/oauthlib/issues/686 and probably some others (https://github.com/oauthlib/oauthlib/pull/640 ?).

### DIFF
--- a/docs/oauth2/grants/custom_grant.rst
+++ b/docs/oauth2/grants/custom_grant.rst
@@ -7,10 +7,9 @@ which is in an early draft, or implement a grant provided by a
 specific OAuth2.0 Authorization Server documentation but not provided
 by oauthlib. For information, any grant types with a clear
 specification can be integrated in oauthlib, just make a PR for that
-!. See :doc:`how to contribute here </contributing>`.
+! See :doc:`how to contribute here </contributing>`.
 
-Please find below an example of how to create a new grant and use it
-in an endpoint:
+Please find how to create a new grant and use it in an endpoint:
 
 .. contents:: Tutorial Contents
     :depth: 3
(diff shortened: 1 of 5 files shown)


docs/oauth2/grants/grants.rst
```diff
@@ -9,23 +9,30 @@ Grant types
...
-OAuth 2, the Client Credentials grant is excellent for embedded services and
-backend applications.
+Grant types are what make OAuth 2 so flexible. The :doc:`Authorization
+Code grant </oauth2/grants/authcode>` is the default for almost all
+Web Applications, the :doc:`Implicit grant </oauth2/grants/implicit>`
+serves less secure applications such as Mobile Applications or
+Single-Page Applications, the :doc:`Client Credentials grant
+</oauth2/grants/credentials>` is excellent for embedded services and
+backend applications. We have also the :doc:`Resource Owner Password
+Credentials grant </oauth2/grants/password>` for legacy applications
```
thedrow: Why legacy applications?
I use OAuth2 all the time for authentication and authorization of my own users.
This requires username and password.

JonathanHuot: That's the usual recommendation from RFC and BCP. See https://tools.ietf.org/html/rfc6749#section-10.7 

>    The resource owner password credentials grant type is often used for
   legacy or migration reasons.

And BCP at https://tools.ietf.org/html/rfc6819#section-4.4.3 . Also, because the general recommendation of the RFC 

>  The authorization server and client SHOULD minimize use of this grant type and utilize other grant types whenever possible. 

Also, I didn't changed much from the previous iteration. If you'd prefer, we could add and emphasize the following sentence from the RFC:

> The credentials should only be used when there is a high
   degree of trust between the resource owner and the client (e.g., the
   client is part of the device operating system or a highly privileged
   application), and when other authorization grant types are not
   available (such as an authorization code).

Thoughts?

thedrow: Yes, that should explain it better.
"Legacy" implies that one should never use the grant for new applications while in fact you should use the Resource Owner Password
Credentials grant when you are the owner of the application and only then.

JonathanHuot:
Any suggestions about the example ? I didn't exactly tried this way, but it should work. I think it can be useful.

thedrow:
I'd try to keep it more realistic to ensure people understand how to implement custom grants.

docs/oauth2/grants/custom_grant.rst
```diff
@@ -7,57 +7,72 @@ which is in an early draft, or implement a grant provided by a
 specific OAuth2.0 Authorization Server documentation but not provided
 by oauthlib. For information, any grant types with a clear
 specification can be integrated in oauthlib, just make a PR for that
-!. See :doc:`how to contribute here </contributing>`.
+! See :doc:`how to contribute here </contributing>`.
```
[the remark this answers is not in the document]
JonathanHuot: Autoformatting, I didn't noticed !

Issue #613, opened 2018-11-20
Unified Custom Validators (hooks) for Grant and Server


JonathanHuot:
**Describe the feature**

Adding Custom Validators to Grants is doable by using the `GrantType` interface (see https://github.com/oauthlib/oauthlib/blob/v2.1.0/oauthlib/oauth2/rfc6749/grant_types/base.py#L49-L56). However, we provide `pre_configured` classes which does not expose "easily" these grant types. Also, we have multiple different kind of "hooks", for `Server.__init__()` we have `token_generator=hook`, `token_expires_in=hook`, `refresh_token_generator=hook`. Then for `GrantTypeBase` have `custom_validators (dict)` (`pre_auth`/`post_auth`/`pre_token`/`post_token`), `register_code_modifier(hook)` and `register_token_modifier(hook)`. 

It is complicated for OAuth2 provider implementation but become worst when using the **OpenID Connect** `openid.pre_configured` which has 8 `GrantTypeBase` and 3 `Dispatchers`, where each `Dispatcher` has two `Grant` with **different** names (e.g. `default_auth_grant` or `default_implicit_grant`).

E.g. Current OAuth2 Provider example:
```python
oauthlib_server = oauth2.Server(..)
oauthlib_server._response_types["token"].custom_validators.pre_auth.append(myhook)
oauthlib_server._response_types["code"].custom_validators.pre_auth.append(myhook)
oauthlib_server._response_types["code"].register_code_modifier(mymodifier)
```

Same hooks with OpenID Provider is just impossible, or will span to more than 20 lines of code.

I open this issue to raise the problem and see what's the solutions we could imagine to progress toward a better place.

**Additional context**

It applies to OAuth2 Provider and OpenID Connect Provider server code.

skion:
My `__init__()` for an OIDC Server is indeed about 150 lines long...

IvanAnishchuk:
Perhaps, instead of relying on any particular endpoint implementation to put those grant types (and token handlers!) into some private dict they could be Server's public attributes from the beginning instead? It shouldn't break any current non-default workflow, I think, and would only require adding a few `self.` references.

Alternatively they could live in some more public dictionary right from the point of their initialization but attributes sounds a bit better to me.

I'd also like `kwargs` not to be ignored in that __init__ and passed to endpoints, tokens, grants as promised. I guess overriding Server class in any more or less complex project is almost unavoidable but it would be good to be able to pass a few extra arguments and register some hooks without having to do that.

[...]

PR #703, opened 2019-10-20
Make grants public attributes of pre_configured server class (#613)

IvanAnishchuk:
To make registering hooks without custom server class easier.

This is the easy part of discussed in #613 , implementing the logic about passing kwargs requires some more discussion IMO.


JonathanHuot:
Looks `token_grant_choice` in OIDC is missing `self.` !

PR #711, opened 2020-01-13
typo in log message for client_credentials

polamayster:
there is no user in client credentials grant type, so I assume it's a typo


oauthlib/oauth2/rfc6749/grant_types/client_credentials.py
```diff
@@ -117,7 +117,7 @@ def validate_token_request(self, request):
         # Ensure client is authorized use of this grant type
         self.validate_grant_type(request)
 
-        log.debug('Authorizing access to user %r.', request.user)
+        log.debug('Authorizing access to client %r.', request.client.client_id)
```
JonathanHuot: client_id could be in `request.client_id` instead. So I'd suggest to move the log.debug below the `request.client_id = ` line.

PR #717, opened 2020-02-18
Replaced mention of flask-oauthlib with authlib

InnovativeInventor:
From the GitHub page, it looks like https://github.com/lepture/flask-oauthlib has been deprecated in favor of https://github.com/lepture/authlib. I replaced the link/mention to maintain consistency and keep the README up to date.


JonathanHuot:
Rejected PR because authlib does not use oauthlib at all.

Issue #722, opened 2020-03-26
Support for RSA-SHA256 and RSA-SHA512 in OAuth 1.0a


hoylen:
**Describe the feature**

Support new OAuth 1.0a _signature methods_ that use SHA-256 and SHA-512 hash algorithms. These would function exactly like the standard RSA-SHA1 _signature method_, but replaces the SHA-1 algorithm with SHA-256 or SHA-512 algorithms.

OAuthLib already implements the HMAC-SHA256 _signature method_, so there is precedence for supporting _signature methods_ beyond those specified by RFC 5849.

**Additional context**

SHA-1 has now been deprecated by NIST and its use discouraged or forbidden, because successful attacks on it have been demonstrated,

However, there are situations where the functionality provided by OAuth 1.0a is still needed, and OAuth 2.0 is not a suitable alternative.

hoylen:
Issue resolved by pull request https://github.com/oauthlib/oauthlib/pull/723.

PR #723, opened 2020-03-26
OAuth 1.0a signature methods: RSA-SHA256, RSA-SHA512 and HMAC-SHA512

hoylen:
Implemented new OAuth 1.0a _signature methods_ that work exactly like RSA-SHA1, but replaces the deprecated SHA-1 algorithm with the more stronger SHA-256 or SHA-512 algorithm. For completeness, also implemented a HMAC-based _signature method_ that also replaces SHA-1 with SHA-512.

This pull request addresses https://github.com/oauthlib/oauthlib/issues/722.

**Overview of the changes**

_Implementation_

The code in _oauthlib/oauth1/rfc5849/signatures.py_ was reorganised so all of the HMAC-based signing functions calls a common `_sign_hmac` function with the hash function to use as a parameter. Similarly, all of the RSA-based signing functions calls a common `_sign_rsa` function. This is better than having lots of duplicated code (with only has one small difference between them) which would be harder to understand and maintain.

This has resulted in some functions becoming deprecated. The old code implemented signing with a pair of functions. For example, `sign_rsa_sha1_with_client` and `sign_rsa_sha1`. The first calls the second. Now with a common `_sign_rsa` function, the first directly calls the new common function - so the second function has been deprecated.

_Tests_

The code in the unit test, _tests/oauth1/rfc5849/test_signatures.py_, has been substantially rewritten to be more clear and complete. It now tests the signature verification functions (it didn't before). It also achieves complete code coverage for _signatures.py_, with the exception of several functions which are now considered deprecated.

_Documentation_

There are some updates to the feature matrix document and the installation instructions document. The installation needed to be corrected to ensure RSA support is properly installed, if RSA-based _signature methods_ are needed.

Some other files were updated so that the documentation built without errors. However, most of the changed are limited to _oauthlib/oauth1/rfc5849/signatures.py_, files that imports and uses it, and the unit tests for it.


docs/feature_matrix.rst
```diff
@@ -1,58 +1,89 @@
...
+
+All standard *signature methods* defined in `RFC 5849`_ *The OAuth 1.0
+Protocol* are supported:
+
+- HMAC-SHA1
+- RSA-SHA1
+- PLAINTEXT
+
+Non-standard *signature methods* that replaces SHA-1 with more
+stronger digest algorithms are also supported:
```
thedrow:
```suggestion
Non-standard *signature methods* that replaces SHA-1 with stronger
digest algorithms are also supported:
```

oauthlib/oauth1/__init__.py
```diff
@@ -7,11 +7,21 @@
 and Server classes.
 """
 from .rfc5849 import Client
-from .rfc5849 import SIGNATURE_HMAC, SIGNATURE_HMAC_SHA1, SIGNATURE_HMAC_SHA256, SIGNATURE_RSA, SIGNATURE_PLAINTEXT
+from .rfc5849 import \
```
thedrow: I'd rather use parenthesis instead of `\` to break lines.

oauthlib/oauth1/rfc5849/endpoints/base.py
```diff
@@ -177,38 +181,65 @@ def _check_mandatory_parameters(self, request):
 
     def _check_signature(self, request, is_token_request=False):
         # ---- RSA Signature verification ----
-        if request.signature_method == SIGNATURE_RSA:
+        if request.signature_method == SIGNATURE_RSA_SHA1 or \
+           request.signature_method == SIGNATURE_RSA_SHA256 or \
+           request.signature_method == SIGNATURE_RSA_SHA512:
```
thedrow:
```suggestion
        if request.signature_method in (SIGNATURE_RSA_SHA1, SIGNATURE_RSA_SHA256, SIGNATURE_RSA_SHA512):
```

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -93,37 +96,40 @@ def signature_base_string(http_method, base_str_uri,
     # 5.  The request parameters as normalized in `Section 3.4.1.3.2`_, after
     #     being encoded (`Section 3.6`).
     #
-    # .. _`Section 3.4.1.3.2`: https://tools.ietf.org/html/rfc5849#section-3.4.1.3.2
+    # .. _`Sec 3.4.1.3.2`: https://tools.ietf.org/html/rfc5849#section-3.4.1.3.2
```
thedrow: Why did you change this?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -134,26 +140,27 @@ def base_string_uri(uri, host=None):
     #
     # .. _`RFC3986`: https://tools.ietf.org/html/rfc3986
 
-    if not scheme or not netloc:
```
thedrow: why is this no longer the case?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -199,77 +226,29 @@ def base_string_uri(uri, host=None):
...
 
     Parameters starting with `oauth_` will be unescaped.
 
     Body parameters must be supplied as a dict, a list of 2-tuples, or a
-    formencoded query string.
+    form encoded query string.
 
     Headers must be supplied as a dict.
 
-    Per `section 3.4.1.3.1`_ of the spec.
```
thedrow: Why did you remove the example?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -280,11 +259,11 @@ def collect_parameters(uri_query='', body=[], headers=None,
     #    `RFC3986, Section 3.4`_.  The query component is parsed into a list
     #    of name/value pairs by treating it as an
     #    "application/x-www-form-urlencoded" string, separating the names
-    #    and values and decoding them as defined by
-    #    `W3C.REC-html40-19980424`_, Section 17.13.4.
+    #    and values and decoding them as defined by W3C.REC-html40-19980424
+    #    `W3C-HTML-4.0`_, Section 17.13.4.
```
thedrow: why did you change this?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -280,11 +259,11 @@ def collect_parameters(uri_query='', body=[], headers=None,
...
     #
-    # .. _`RFC3986, Section 3.4`: https://tools.ietf.org/html/rfc3986#section-3.4
-    # .. _`W3C.REC-html40-19980424`: https://tools.ietf.org/html/rfc5849#ref-W3C.REC-html40-19980424
+    # .. _`RFC3986, Sec 3.4`: https://tools.ietf.org/html/rfc3986#section-3.4
+    # .. _`W3C-HTML-4.0`: https://www.w3.org/TR/1998/REC-html40-19980424/
```
thedrow: why did you change this?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -307,12 +286,12 @@ def collect_parameters(uri_query='', body=[], headers=None,
     #
     #     *  The entity-body follows the encoding requirements of the
     #        "application/x-www-form-urlencoded" content-type as defined by
-    #        `W3C.REC-html40-19980424`_.
+    #        W3C.REC-html40-19980424 `W3C-HTML-4.0`_.
```
thedrow: why did you change this?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -307,12 +286,12 @@ def collect_parameters(uri_query='', body=[], headers=None,
...
 
     #     *  The HTTP request entity-header includes the "Content-Type"
     #        header field set to "application/x-www-form-urlencoded".
     #
-    # .._`W3C.REC-html40-19980424`: https://tools.ietf.org/html/rfc5849#ref-W3C.REC-html40-19980424
+    # .. _`W3C-HTML-4.0`: https://www.w3.org/TR/1998/REC-html40-19980424/
```
thedrow: why did you change this?

oauthlib/oauth1/rfc5849/signature.py
```diff
@@ -334,75 +313,17 @@ def collect_parameters(uri_query='', body=[], headers=None,
...
-def normalize_parameters(params):
-    """**Parameters Normalization**
-    Per `section 3.4.1.3.2`_ of the spec.
-
-    For example, the list of parameters from the previous section would
-    be normalized as follows:
-
-    Encoded::
-
-    +------------------------+------------------+
```
thedrow: Why did you remove the example?

requirements.txt
```diff
@@ -1,3 +1,3 @@
-pyjwt==1.6.0
-blinker==1.4
+pyjwt>=1.6.0
+blinker>=1.4
```
thedrow: Why did you change this?

setup.py
```diff
@@ -18,9 +18,9 @@ def fread(fn):
         return f.read()
 
 
-rsa_require = ['cryptography']
-signedtoken_require = ['cryptography', 'pyjwt>=1.0.0']
-signals_require = ['blinker']
+rsa_require = ['cryptography>=1.4.0']
+signedtoken_require = ['cryptography>=1.4.0', 'pyjwt>=1.6.0']
+signals_require = ['blinker>=1.4.0']
```
thedrow: Why did you change this?
Is it necessary for the feature to work?

oauthlib/oauth1/rfc5849/signature.py
hoylen: To satisfy PEP-8 maximum line length rule.

oauthlib/oauth1/rfc5849/signature.py
hoylen: Test for missing `scheme` error is still done in the new line 143. But the `netloc` obtained from _urlparse_ **can** be empty if an overriding value is provided in the `host` parameter. It is only an error if **both** _netloc_ is empty and the _host_ is empty: see new line 200.

oauthlib/oauth1/rfc5849/signature.py
hoylen: It added little value, because:
1. The reader can easily load up the RFC and read it directly.
2. The inline comments for the algorithm are already present.
3. I doubt readers will scroll up and down between the code and example. And if they did, it would be easier done in a separate window (point 1).
3. If the whole example was to be added as a comment, it would make more sense to copy it into the unit test _tests/oauth1/rfc5849/test_signatures.py_ which uses the example for its testing. But again, duplicating the RFC verbatim doesn't add much value.

oauthlib/oauth1/rfc5849/signature.py
hoylen: The shorter reference identifier is more easily recognisable as "HTML 4.0". The "-40-19980424" makes it harder for the human reader and adds little value in this context. Also, to satisfy PEP-8 maximum line length rule.

oauthlib/oauth1/rfc5849/signature.py
hoylen: The shorter reference identifier is more easily recognisable as "HTML 4.0". The "-40-19980424" makes it harder for the human reader and adds little value in this context. Also, to satisfy PEP-8 maximum line length rule.

oauthlib/oauth1/rfc5849/signature.py
hoylen: The shorter reference identifier is more easily recognisable as "HTML 4.0". The "-40-19980424" makes it harder for the human reader and adds little value in this context. Also, to satisfy PEP-8 maximum line length rule.

oauthlib/oauth1/rfc5849/signature.py
hoylen: The shorter reference identifier is more easily recognisable as "HTML 4.0". The "-40-19980424" makes it harder for the human reader and adds little value in this context. Also, to satisfy PEP-8 maximum line length rule.

oauthlib/oauth1/rfc5849/signature.py
hoylen: It added little value, because:

The reader can easily load up the RFC and read it directly.
The inline comments for the algorithm are already present.
I doubt readers will scroll up and down between the code and example. And if they did, it would be easier done in a separate window (point 1).
If the whole example was to be added as a comment, it would make more sense to copy it into the unit test tests/oauth1/rfc5849/test_signatures.py which uses the example for its testing. But again, duplicating the RFC verbatim doesn't add much value.

requirements.txt
hoylen: To align it with the values inside _setup.py_ -- is this the wrong way to do it? The _setup.py_ had "pyjwt>=1.0.0", but _requirements.txt_ had a different "pyjwt==1.6.0" So for consistency, made both reference the same v1.6.0 and made them both ">=". For consistency, also added a ">=" for _blinker_, picking the most recent release.

setup.py
hoylen: Seemed dangerous not to specify a version number for the _cryptography_ package, especially when there are security issues in older versions. Have only tested with the latest v2.8. Happy to remove the version numbers if that is the approach _oauthlib_ wants to take.

oauthlib/oauth1/rfc5849/signature.py
hoylen: I'm using PyCharm IDE, so fixing the PEP-8 warnings means the file comes up with no detected issues. Otherwise, the warnings are noise that one has to ignore, and one might miss a real error/warning in amongst them.

requirements.txt
hoylen: The Python Packaging Authority [user guide](https://packaging.python.org/discussions/install-requires-vs-requirements/) and this article on [setup.py vs requirements.txt](https://caremad.io/posts/2013/07/setup-vs-requirement/) says the current best practice is to have ">=" in the _setup.py_ and for the _requirements.txt_ to reference _setup.py_ or only have "==". But that makes an assumption on how the recipient intends to use those files. Maybe there should only be a _setup.py_ and no _requirements.txt_ file?

Fixing this sounds like a new issue in itself. Perhaps accept the changes that are in this pull request since it is better than it was (having them consistent is better than having them inconsistent), and a new issue should be opened to determine an even better fix.

hoylen: Opened issue https://github.com/oauthlib/oauthlib/issues/724 and resolving this conversation.

setup.py
hoylen: Can be addressed when issue https://github.com/oauthlib/oauthlib/issues/724 is addressed.

hoylen:
@thedrow I have addressed the review comments you made on 2020-03-29, but this PR still claims there are "Changes requested" (in big bold red text). This seems to be a bug/feature in GitHub, and you'll need to do something to move the PR forward. Thanks.

hoylen:
@auvipy Merge conflict fixed. Thanks.

The dependency in master on pyjwt had been bumped up from 1.6.0 to 1.7.1. Change merged.

hoylen:
@auvipy I have addressed the review comments you made on 2020-04-09, but this PR still claims there are "Changes requested" (in big bold red text). This seems to be a bug/feature in GitHub, and you'll need to do something to move the PR forward. Thanks.

docs/feature_matrix.rst
```diff
@@ -1,58 +1,89 @@
...
 - OpenID Discovery
 - OpenID Session Management
 - ...and more
 
-Supported platforms
--------------------
+Platforms
+---------
+
+OAuthLib is mainly developed andtested on 64-bit Linux. It works on
```
thedrow, line 64:
```suggestion
OAuthLib is mainly developed and tested on 64-bit Linux. It works on
```

thedrow:
I'm going to wait for another day to see if someone else would like to review this PR.
If not, I'll merge it.

hoylen:
Thanks Omer. Much appreciated.

PR #729, opened 2020-04-19
OAuth2 `xxxApplicationClient` now correctly use the default `scope` provided in constructor

smarie:
Fixes #728

PR #726, opened 2020-04-09
OAuth2 `LegacyApplicationClient.prepare_request_body` now correctly uses the default `scope` provided in constructor

smarie:
Fixes #725

PR #731, opened 2020-04-19
Fix OAuth2 base `Client` inconsistent `scope` handling

smarie:
This is the final step after PRs #682 #726 and #729 : I found out that even the base `Client` has inconsistent management of the `scope` attribute.

 - in `prepare_authorization_request` and `prepare_refresh_token_request` if a `scope` is provided, it is assigned to `self` forever, overriding any scope provided in the constructor

 - in `parse_request_body_response` and `prepare_refresh_body`, if no scope is provided the one from the constructor is not used !

This PR is an attempt to uniformize everything, but I think that we should ask ourselves the question: will it have side-effects ? (are there enough tests as of today to ensure non-regression)


smarie:
Thanks @JonathanHuot !

PR #735, opened 2020-05-10
Py38

auvipy:


auvipy:
will rebase

tox.ini
```diff
@@ -11,7 +11,7 @@ commands=
 # tox -e docs to mimick readthedocs build.
 # as of today, RTD is using python3.6 and doesn't run "setup.py install"
 [testenv:docs]
-basepython=python3.6
+basepython=python3.8
```
JonathanHuot, line 14: have you verified RTD is using python3.8 and no longer python3.6 ? when the comment above was written, it was 3.6, at least :)

auvipy: i think need to revert this

JonathanHuot: Can you rever tit ? Then we merge the PR

auvipy: this can be squash merge now

PR #743, opened 2020-11-22
Added pkce on client side for authorization grant flow. Test cases added

amans330:
Hi, Please review my pull request for PKCE implementation on the client-side. Do let me know if any changes are required.


amans330:
@auvipy I noticed a build fail. Though it doesn't seem to be caused by my code change. Anything I need to do about this?

oauthlib/oauth2/rfc6749/clients/web_application.py
```diff
@@ -87,9 +97,10 @@ def prepare_request_uri(self, uri, redirect_uri=None, scope=None,
         """
         scope = self.scope if scope is None else scope
         return prepare_grant_uri(uri, self.client_id, 'code',
-                                 redirect_uri=redirect_uri, scope=scope, state=state, **kwargs)
+                                 redirect_uri=redirect_uri, scope=scope, state=state, code_challenge=code_challenge,
+                                 code_challenge_method=code_challenge_method, **kwargs)
 
-    def prepare_request_body(self, code=None, redirect_uri=None, body='',
+    def prepare_request_body(self, code=None, redirect_uri=None, code_verifier=None, body='',
```
JonathanHuot, line 103: Generally, it is better to add new arguments after all existing kwargs, in case someone use them as args; to preserve retrocompatibility.
Would you mind moving `code_verifier` just before **kwargs ?

oauthlib/oauth2/rfc6749/clients/web_application.py
```diff
@@ -153,8 +169,8 @@ def prepare_request_body(self, code=None, redirect_uri=None, body='',
 
         kwargs['client_id'] = self.client_id
         kwargs['include_client_id'] = include_client_id
-        return prepare_token_request(self.grant_type, code=code, body=body,
-                                     redirect_uri=redirect_uri, **kwargs)
+        return prepare_token_request(self.grant_type, code=code, code_verifier=code_verifier,
+                                     body=body, redirect_uri=redirect_uri, **kwargs)
```
JonathanHuot, line 173: Same

oauthlib/oauth2/rfc6749/parameters.py
```diff
@@ -81,7 +89,7 @@ def prepare_grant_uri(uri, client_id, response_type, redirect_uri=None,
     return add_params_to_uri(uri, params)
 
 
-def prepare_token_request(grant_type, body='', include_client_id=True, **kwargs):
+def prepare_token_request(grant_type, code_verifier=None, body='', include_client_id=True, **kwargs):
```
JonathanHuot, line 92: same

JonathanHuot (changes requested):

thedrow (approved):
LGTM. Please address code review comments.

amans330:
@thedrow will finish it by eod today. Just starting a new job today!

amans330:
Hi @JonathanHuot I made the changes based on the above comments. Please review.

auvipy (approved):

amans330:
@JonathanHuot This PR is awaiting review from you.

JonathanHuot (changes requested):
Great PR so far, I have added minor comments inline.

Also, I think it is worth to put additional python functions to compute code_challenge_value and code_verifier. The oauthlib framework still needs to be open to any custom values, but at the same time, by default, it should prepare the ground for the implementor. For example, see the RFC:

```
   The client first creates a code verifier, "code_verifier", for each
   OAuth 2.0 [RFC6749] Authorization Request, in the following manner:

   code_verifier = high-entropy cryptographic random STRING using the
   unreserved characters [A-Z] / [a-z] / [0-9] / "-" / "." / "_" / "~"
   from Section 2.3 of [RFC3986], with a minimum length of 43 characters
   and a maximum length of 128 characters.

   ABNF for "code_verifier" is as follows.

   code-verifier = 43*128unreserved
   unreserved = ALPHA / DIGIT / "-" / "." / "_" / "~"
   ALPHA = %x41-5A / %x61-7A
   DIGIT = %x30-39

   NOTE: The code verifier SHOULD have enough entropy to make it
   impractical to guess the value.  It is RECOMMENDED that the output of
   a suitable random number generator be used to create a 32-octet
   sequence.  The octet sequence is then base64url-encoded to produce a
   43-octet URL safe string to use as the code verifier.
```

and
```
   The client then creates a code challenge derived from the code
   verifier by using one of the following transformations on the code
   verifier:

   plain
      code_challenge = code_verifier

   S256
      code_challenge = BASE64URL-ENCODE(SHA256(ASCII(code_verifier)))

   If the client is capable of using "S256", it MUST use "S256", as
   "S256" is Mandatory To Implement (MTI) on the server.  Clients are
   permitted to use "plain" only if they cannot support "S256" for some
   technical reason and know via out-of-band configuration that the
   server supports "plain".
```

By doing this in oauthlib side, it will prevent implementor to do mistake during the implementation and will be very easy to implement it right.
Thoughts?

Issue #746, opened 2021-02-01
OIDC Hybrid Flow - nonce not passed to `add_id_token`


tevansuk:
When using OIDC Hybrid Flow, the nonce not passed when calling `GrantTypeBase.add_id_token`, despite it being present in `request.nonce`.

In the OIDC implicit flow, `add_id_token` is overridden to pull the nonce from `request.nonce`. In OIDC Authorization Code flow, we pull the nonce from the grant, but in Hybrid Flow nothing is done to handle the nonce, so it is omitted from the ID token. Putting the nonce in the ID token is REQUIRED for Hybrid Flow ([3.3.2.11](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken)).

~When retrieving the token from both the Authorization Endpoint and the Token Endpoint, the contents of the token should be the same - with some differences ([3.3.3.6](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken2)). The key phrase for me is `All Claims about the Authentication event present in either SHOULD be present in both.`~

~The nonce is a claim about the authentication event and must be present in the ID token returned from the authentication endpoint, so it should also be present in the id token from the token endpoint. This means that when handling a auth code token request in the hybrid flow, we should also be calling `self.request_validator.get_authorization_code_nonce` to retrieve any nonce from the grant.~

EDIT: of course, when handling generating the ID token from the code, it doesn't matter that the code originally came from the hybrid flow, the token endpoint is handled by `AuthorizationCodeGrant`, which will inject the nonce just fine as long as it was correctly saved/restored along with the code.

**How to reproduce**

I'll add some tests to demonstrate.

**Expected behavior**

A nonce passed in to the authentication endpoint in hybrid flow should be present in the ID token from the authentication endpoint and the token endpoint.

**Additional context**

- Are you using OAuth1, OAuth2 or OIDC?
- OIDC
- Are you writing client or server side code?
- server side
- If client, what provider are you connecting to?
- N/A
- Are you using a downstream library, such as `requests-oauthlib`, `django-oauth-toolkit`, ...?
- Adding OIDC support to `django-oauth-toolkit`

PR #747, opened 2021-02-01
Use request.nonce when generating hybrid id token

tevansuk:
Like with the implicit grant, we need to override `add_id_token` to pass the nonce from the current request to `GrantBase.add_id_token` in order for the ID token to have the correct nonce.

Add test that the nonce is in ID token from hybrid OIDC flow.

Fixes: #746

### DIFF
--- a/oauthlib/openid/connect/core/grant_types/hybrid.py
+++ b/oauthlib/openid/connect/core/grant_types/hybrid.py
@@ -35,6 +35,9 @@ class HybridGrant(GrantTypeBase):
         self.register_code_modifier(self.add_id_token)
         self.register_token_modifier(self.add_id_token)
 
+    def add_id_token(self, token, token_handler, request):
+        return super().add_id_token(token, token_handler, request, nonce=request.nonce)
+
     def openid_authorization_validator(self, request):
         """Additional validation when following the Authorization Code flow.
         """
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,3 +29,4 @@ Brendan McCollam
 Jonathan Huot
 Pieter Ennes
 Olaf Conradi
+Tom Evans
--- a/tests/openid/connect/core/grant_types/test_hybrid.py
+++ b/tests/openid/connect/core/grant_types/test_hybrid.py
@@ -67,6 +67,15 @@ class OpenIDHybridCodeIdTokenTest(OpenIDAuthCodeTest):
         self.assertIsNone(b)
         self.assertEqual(s, 302)
 
+    def test_id_token_contains_nonce(self):
+        token = {}
+        self.mock_validator.get_id_token.side_effect = None
+        self.mock_validator.get_id_token.return_value = None
+        token = self.auth.add_id_token(token, None, self.request)
+        assert self.mock_validator.finalize_id_token.call_count == 1
+        claims = self.mock_validator.finalize_id_token.call_args[0][0]
+        assert "nonce" in claims
+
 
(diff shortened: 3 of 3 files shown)


tevansuk:
I see two ways of fixing this. The first is what is in the PR, which follows how this is done in OIDC auth code flow and OIDC implicit flow, to override `add_id_token` and inject the `nonce`.

The second is to do away with all these overrides (and `RequestValidator.get_authorization_code_nonce`, and have `GrantTypeBase.add_id_token` to just look at `request.nonce`. For auth code flow, this would involve a change in behaviour for OIDC providers to require  `RequestValidator.validate_code` to set `request.nonce` when validating the code. This would be a backwards incompatible change however, and so I went with the first option.

oauthlib/openid/connect/core/grant_types/hybrid.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, request_validator=None, **kwargs):
         self.register_code_modifier(self.add_id_token)
         self.register_token_modifier(self.add_id_token)
 
+    def add_id_token(self, token, token_handler, request):
+        return super().add_id_token(token, token_handler, request, nonce=request.nonce)
```
auvipy, line 39: is this nonce related to this? https://stackoverflow.com/questions/3690738/need-help-understanding-nonce

tevansuk: Yep, or better description is [in the spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest). Its used to prevent replay attacks, the client generates a random nonce and stores it in a session, and then compares the nonce in the response with the nonce in the session. If it is missing either in the session or the response, or does not match, the response is invalid.

Its a required parameter for the implicit flow, a sometimes required parameter for the hybrid flow and an optional parameter for authorization code flow.

Handling it, however, is not optional - if a nonce is presented in the authorization endpoint, it must be in the ID token or a client should reject the response as invalid.

JonathanHuot:
Hi @tevansuk , good catch ! However note that nonce is optional for "code token" in hybrid mode, as highlighted in https://github.com/oauthlib/oauthlib/blob/d54965b86ce4ede956db70baff0b3d5e9182a007/oauthlib/openid/connect/core/grant_types/base.py#L115-L123 ; however your code seems anyway using the right approach (using request.nonce will work even without value).

JonathanHuot (commented):
LGTM

tevansuk:
Thanks for the review!

I'm not sure that text is correct by the way. The spec says that it is an optional authentication request parameter in certain flows, and required in other flows, but it *can* be specified in all of them, and if it is used, the server *must* include it in the ID token, and clients *must* verify it  - [section 2](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) (ID Token)

> If present in the ID Token, Clients MUST verify that the nonce Claim Value is equal to the value of the nonce parameter sent in the Authentication Request. If present in the Authentication Request, Authorization Servers MUST include a nonce Claim in the ID Token with the Claim Value being the nonce value sent in the Authentication Request. Authorization Servers SHOULD perform no other processing on nonce values used.

Then in [section 3.3.2.11](https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken) (Hybrid Flow - ID Token):

> nonce
>     Use of the nonce Claim is REQUIRED for this flow.

For "code token" of course, the id token is retrieved by the auth code flow grant handler, which handles it the same as a standard auth code, so it did already work correctly for that case!

auvipy (approved):

PR #752, opened 2021-03-22
Oidc refresh

### LINKED ISSUE #751: Refresh id tokens
**Describe the feature**
It is not possible with the current implementation to issue ID tokens on refresh.

**Additional context**

The refresh token modifiers (https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py#L66) take only 1 argument while the authorization code modifiers take 3 (https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/grant_types/authorization_code.py#L308). 

The modifier function `add_id_token` used to add id tokens in the OIDC token response is not compatible with the refresh token's modifier interface. 

I see no reason to have that behavior. If there are no objections I can go ahead and create a PR that will harmonize the modifier arguments and add ID tokens to the refresh token responses for OIDC out of the box.

nsklikas:
Fixes #751 

This PR adds support for refreshing ID Tokens.

This PR breaks backwards compatibility with `RefreshTokenGrant` modifiers. If this is a problem, I could change the `RefreshTokenGrant` modifiers only for OIDC in order to remain backwards compatible.

### DIFF
--- a/oauthlib/openid/connect/core/grant_types/refresh_token.py
+++ b/oauthlib/openid/connect/core/grant_types/refresh_token.py
@@ -1,0 +1,36 @@
+"""
+oauthlib.openid.connect.core.grant_types
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+"""
+import logging
+
+from oauthlib.oauth2.rfc6749.grant_types.refresh_token import (
+    RefreshTokenGrant as OAuth2RefreshTokenGrant,
+)
+
+from .base import GrantTypeBase
+
+log = logging.getLogger(__name__)
+
+
+class RefreshTokenGrant(GrantTypeBase):
+
+    def __init__(self, refresh_id_token=True, request_validator=None, **kwargs):
+        self.refresh_id_token = refresh_id_token
+        self.proxy_target = OAuth2RefreshTokenGrant(
+            request_validator=request_validator, **kwargs)
+        self.register_token_modifier(self.add_id_token)
+
+    def add_id_token(self, token, token_handler, request):
+        """
+        Construct an initial version of id_token, and let the
+        request_validator sign or encrypt it.
+
+        The authorization_code version of this method is used to
+        retrieve the nonce accordingly to the code storage.
+        """
+        # Treat it as normal OAuth 2 auth code request if openid is not present
+        if not self.refresh_id_token:
+            return token
+
+        return super().add_id_token(token, token_handler, request)
--- a/oauthlib/openid/connect/core/grant_types/refresh_token.py
+++ b/oauthlib/openid/connect/core/grant_types/refresh_token.py
@@ -15,8 +15,7 @@ log = logging.getLogger(__name__)
 
 class RefreshTokenGrant(GrantTypeBase):
 
-    def __init__(self, refresh_id_token=True, request_validator=None, **kwargs):
-        self.refresh_id_token = refresh_id_token
+    def __init__(self, request_validator=None, **kwargs):
         self.proxy_target = OAuth2RefreshTokenGrant(
             request_validator=request_validator, **kwargs)
         self.register_token_modifier(self.add_id_token)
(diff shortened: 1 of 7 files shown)


auvipy (approved):

oauthlib/openid/connect/core/grant_types/refresh_token.py
```diff
@@ -0,0 +1,36 @@
...
+    def add_id_token(self, token, token_handler, request):
+        """
+        Construct an initial version of id_token, and let the
+        request_validator sign or encrypt it.
+
+        The authorization_code version of this method is used to
+        retrieve the nonce accordingly to the code storage.
+        """
+        # Treat it as normal OAuth 2 auth code request if openid is not present
+        if not self.refresh_id_token:
```
JonathanHuot, line 33: I don't see how do we set this boolean ? in the unit test it is done explicitely, but actually it should be done accordingly to if the scope includes "openid", right ?

JonathanHuot (commented):
Can you check & update the documentation generated for oidc please ?

tox -e docs 

https://oauthlib.readthedocs.io/en/latest/oauth2/oidc.html#

oauthlib/openid/connect/core/grant_types/refresh_token.py
nsklikas: It is set during the initialization of the class (https://github.com/oauthlib/oauthlib/pull/752/files#diff-a9b30143f6cf03270ce00f8311710d82eaad0072eb5c01dd3aef15412c70ff68R19) and defaults to `True` so that the user can omit refreshing id tokens if he wishes to

zzzevaka:
Hello! I am making the same functional in my own project and faced with a little problem.

An attribute request.client_id is used in add_id_token, but request.client_id is not set if using client_secret_basic authentication rather than client_secret_post.

I suggest to make modification in add_id_token in this PR:

```python
- id_token['aud'] = request.client_id
+ id_token['aud'] = request.client_id or request.client.client_id
```

nsklikas:
Whether request has the field `client_id` or not, depends on your validator.

I don't think this is the right PR to make this change, unless @JonathanHuot thinks otherwise.

nsklikas:
You are correct, this seems to be a bug. But I don't know if this is the right PR to fix this, unless @JonathanHuot thinks otherwise

nsklikas:
Any news on this?

oauthlib/openid/connect/core/grant_types/refresh_token.py
JonathanHuot: I see a problem if the Token Endpoint is used for both OAuth2.0 refresh and OIDC refresh. When this grant receives a refresh request, it has no dynamic choice to returns an id_token or not.

JonathanHuot:
Hi @zzzevaka , a new PR seems more appropriate for your fix. 
@nsklikas , changes are required to support both oauth2 & OIDC provider ; can you have a look please ?
Thanks!

oauthlib/openid/connect/core/grant_types/refresh_token.py
nsklikas: What kind of dynamic logic do you have in mind?

If this grant is used for both OAuth2 and OIDC and you want to issue ID Tokens on refresh then this param is enough. `add_id_token` adds an ID Token in the response only if the `openid` scope was requested.

In addition if the client removes the `openid` scope in this refresh request then an ID Token will not be issued, which I think is the correct behaviour.

Do you think I should add more tests to validate this?

nsklikas:
It's not clear to me what changes are required, can you please elaborate?

JonathanHuot:
Depending the AS implementation, the id_token can be refreshed :
1) never 
2) always
3) depending id_token parameter, like auth_time, or exp of id_token

I see that 1) and 2) can be implemented with the boolean, but not 3). I was wondering if it is possible to implement to support 3) as well.
Maybe if we add a RequestValidator method?
Any thoughts anyone ?

nsklikas:
Yea, that makes sense. Moving it to the validator provides more flexibility.

I removed the class variable

nsklikas:
I think it should be fine now.

nsklikas:
Any changes needed?

JonathanHuot:
LGTM Thanks for the feature!

PR #753, opened 2021-04-01
Use better regex for IPv6 to allow a lot more valid IPv6 addresses

pauldekkers:
The current IPv6address regular expression leaves out a lot of IPv6 addresses, among which the loopback address `::1` which is important for Loopback Interface Redirection (as described in RFC8252, https://tools.ietf.org/html/rfc8252#section-7.3). Also shorter addresses like fd01::1 or anything between this and a full length address seemed to be left out.

This is a much shorter regular expression, and a suggestion for one that does match a lot of valid IPv6 addresses (if done by regular expressions - an alternative would be the use of socket or maybe IPAddr).


auvipy (changes requested):
thanks for the PR, can you also update some unit tests?

pauldekkers:
Hi! Hope this is what you had in mind - there were no unit tests at all for uri_validate or is_absolute_uri() so I tested for some cases, among which the ones that failed before the a046668

tests/test_uri_validate.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
auvipy, line 1: # -*- coding: utf-8 -*- this line is not needed

pauldekkers: Thanks, I removed it.

auvipy (approved):

pauldekkers:
Hi; it looks like this PR is stuck on review, @auvipy @JonathanHuot?

PR #757, opened 2021-04-27
Properly handle prompt=none

### LINKED ISSUE #756: prompt=none is not handled correctly
**Describe the bug**
The behavior  for handling `prompt=none` described in the OIDC spec (https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) is:
- If `prompt=none`, return error it user isn't authenticated
- If `prompt=none + something else` then return an error.

The current implementation raises an `OIDCNoPrompt` exception whenever we get `prompt=none`, the code to properly handle the exists in https://github.com/oauthlib/oauthlib/blob/master/oauthlib/openid/connect/core/grant_types/base.py#L302

**Expected behavior**

The functions `validate_silent_login` and `validate_silent_authorization` must be called and the proper errors must be raised

nsklikas:
Closes #756


auvipy (approved):

wiliamsouza (approved):
LGTM

Issue #759, opened 2021-05-26
OIDC misinterprets "Authorization: Basic" as Bearer


n2ygk:
**Describe the bug**

An "Authorization: Basic ..." header is misinterpreted as "Authorization: Bearer ..." in: https://github.com/oauthlib/oauthlib/blob/2b8a44855a51ad5a5b0c348a08c2564a2e197ea2/oauthlib/openid/connect/core/tokens.py#L37-L44

This is handled correctly (by checking for the type) in: https://github.com/oauthlib/oauthlib/blob/2b8a44855a51ad5a5b0c348a08c2564a2e197ea2/oauthlib/oauth2/rfc6749/tokens.py#L249-L254

**How to reproduce**

See https://github.com/jazzband/django-oauth-toolkit/issues/964

**Expected behavior**

`token == None` should be returned.

**Additional context**

Please provide any further context here.

- Are you using OAuth1, OAuth2 or OIDC? **OIDC**
- Are you writing client or server side code? **Server side**
- If client, what provider are you connecting to? **N/A**
- Are you using a downstream library, such as `requests-oauthlib`, `django-oauth-toolkit`, ...? **django-oauth-toolkit**

PR #760, opened 2021-05-26
Issue759/check authz type

n2ygk:
Fixes: #759 

Fixes the case where an Authorization header that is **not** a Bearer is misinterpreted as a Bearer -- and misaligned as well, if, for instance, the header is a Basic.

### DIFF
--- a/oauthlib/openid/connect/core/tokens.py
+++ b/oauthlib/openid/connect/core/tokens.py
@@ -37,7 +37,9 @@ class JWTToken(TokenBase):
     def validate_request(self, request):
         token = None
         if 'Authorization' in request.headers:
-            token = request.headers.get('Authorization')[7:]
+            split_header = request.headers.get('Authorization').split()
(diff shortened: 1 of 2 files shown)


oauthlib/openid/connect/core/tokens.py
```diff
@@ -37,15 +37,18 @@ def create_token(self, request, refresh_token=False):
...
         return self.request_validator.validate_jwt_bearer_token(
             token, request.scopes, request)
 
     def estimate_type(self, request):
-        token = request.headers.get('Authorization', '')[7:]
-        if token.startswith('ey') and token.count('.') in (2, 4):
-            return 10
-        else:
-            return 0
+        split_header = request.headers.get('Authorization').split()
```
JonathanHuot, line 49: same here ?

JonathanHuot (changes requested):
Good catch, i have added a couple of simplification comments; great job still

JonathanHuot (approved):
LGTM

PR #761, opened 2021-05-26
Remove typo in TokenEndpoint docstring

bellawoo:
Because otherwise I would continue to stare at the typo on Read the Docs. 😅


JonathanHuot (approved):
Thanks for the fix :)

PR #767, opened 2021-07-01
parameters.py - Expanding Error Code Catching

satiowadahc:
One public API I use returns errorCode: <error>. This should now catch a wider variety of errors.

### DIFF
--- a/oauthlib/oauth2/rfc6749/parameters.py
+++ b/oauthlib/oauth2/rfc6749/parameters.py
@@ -261,8 +261,9 @@ def parse_authorization_code_response(uri, state=None):
     if state and params.get('state', None) != state:
         raise MismatchingStateError()
 
-    if 'error' in params:
-        raise_from_error(params.get('error'), params)
+    for key in params.keys():
+        if 'error' in key.lower():
+            raise_from_error(params.get(key), params)
 
     if not 'code' in params:
         raise MissingCodeError("Missing code parameter in response.")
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
would be great if you also update the unit tests

thedrow (changes requested):
Could you please provide an example where this is useful?

satiowadahc:
If I have time this week I'll investigate your testing code.

From Autodesk API an example error is: 
```json
{'developerMessage': 'The required parameter(s) client_id not present in the request', 'errorCode': 'AUTH-008', 'more info': 'https://forge.autodesk.com/en/docs/oauth/v2/developers_guide/error_handling/'}
```
So in parameters.py:
```python
   if 'error' in params:
       raise_from_error(params.get('error'), params)
```
'error' is not in `params`. But it is in an `params` key. Likewise if an API uses 'ERROR' instead of error. it would never be caught as 'error' != 'ERROR'.

satiowadahc:
Or rather I had a few moments now. Note I'm editing on GitHub and haven't cloned locally yet. So this might take a commit or two,

thedrow (approved):

JonathanHuot:
Hi,
oauthlib by-default only allows the RFC, however we have several points of extensions/customization possible to better fit to the reality.

I think the PR is too flexible as it does not enforce the list of field names. While the lowercase check _might_ be a good idea (however a CaseInsensitiveDict would be easier to read?), the string search `"error" in <string>` will be a source of errors (`acceptError`matches).

JonathanHuot (changes requested):
We need to enforce the list of field names.
Change lowercase into a CaseInsensitiveDict.
The string search `"error" in <string>` cannot work (`acceptError` matches).

satiowadahc:
So if we had a predefined list of error codes? This way it follows RFC but can have added cases for API's that don't follow "standards"

```python
self.reportable_errors = ["error", "Error"]

def get_errors(self)
    return self.reportable_errors
    
def add_error_code(self, err)
     self.reportable_errors.append(err)

.
.
.   
for key in params.keys():
    if key in self.reportable_errors:
        raise_from_error(params.get(key), params)
.
.
.
```

JonathanHuot:
Yes, that could work. However, I think the default's list should be what is in the RFC: only includes `error` ?

PR #766, opened 2021-06-03
Prepare 3.2.0 release

JonathanHuot:
The PR is mainly to test travis-ci.com migration.
GitHub webhook has https://notify.travis-ci.org renamed into https://notify.travis-ci.com

PR #777, opened 2021-10-02
Remove upper bound on cryptography version

riconnon:
Cryptography has adopted a firefox-style versioning system where new
feature releases always have new major versions even if they don't have
backwards incompatible changes. This means that an upper bound on the
dependency does not make sense.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 pyjwt>=2.0.0,<3
 blinker==1.4
-cryptography>=3.0.0,<4
+cryptography>=3.0.0
--- a/setup.py
+++ b/setup.py
@@ -16,8 +16,8 @@ def fread(fn):
         return f.read()
 
 
-rsa_require = ['cryptography>=3.0.0,<4']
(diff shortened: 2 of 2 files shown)


auvipy:
what if version 4 break compatibility?

auvipy (commented):
what if version 4 break compatibility?

riconnon:
@auvipy 
the cryptography project never used semver. They made no guarantee of compatability in "minor" versions, only "bugfix", so for example 3.5.0 was not guaranteed to be compatible with 3.4.x
cryptography has just released the latest version 35.0.0 (which you may have expected to be 3.5.0) which is "mostly" compatible with 3.4.x except for a couple of things relating to x509 parsing

Full details here: https://cryptography.io/en/latest/api-stability/#version-35-0-0

auvipy (approved):

riconnon:
@auvipy 
This will likely start to present a fairly major inconvenience to your users once other dependencies start wanting the newer `cryptography` releases.
What's your typical release cycle? When could we expect a release version including this change?

graingert:
I think blocked on https://github.com/oauthlib/oauthlib/pull/766 ?

PR #797, opened 2022-01-08
GitHub Action to lint Python code

cclauss:
Test results: https://github.com/cclauss/oauthlib/actions

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,25 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: black --check . || true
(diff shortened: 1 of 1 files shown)


JonathanHuot:
It looks interesting but while a lot of files are updated I was not able to see the actual diff ?

cclauss:
https://github.com/oauthlib/oauthlib/pull/797/files is the diff... One file changed.

JonathanHuot:
As long as it contains `|| true`, I don't see a problem merging it. And I hope it will encourage others to improve some lint' next.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit black codespell flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
```
JonathanHuot, line 12: Should we reuse the maximum in `tox` as well ? for example, `tox -e bandit` is running `bandit -b bandit.json -r oauthlib` and this github step runs `bandit --recursive --skip B101 . || true`. 
Would imagine better for developers to run those checks from their machines directly instead of from github actions ? thoughts?

cclauss: pre-commit is far cooler than tox and it can be run locally and in CI.

JonathanHuot (approved):
LGTM

PR #803, opened 2022-02-04
Allow non-HTTPS issuer when OAUTHLIB_INSECURE_TRANSPORT.

luhn:
Use `utils.is_secure_transport` to validate issuer, that way `OAUTHLIB_INSECURE_TRANSPORT` is taken into account.

### DIFF
--- a/oauthlib/oauth2/rfc6749/endpoints/metadata.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/metadata.py
@@ -10,13 +10,12 @@ import copy
 import json
 import logging
 
-from .. import grant_types
+from .. import grant_types, utils
 from .authorization import AuthorizationEndpoint
 from .base import BaseEndpoint, catch_errors_and_unavailability
 from .introspect import IntrospectEndpoint
 from .revocation import RevocationEndpoint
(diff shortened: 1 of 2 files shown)


oauthlib/oauth2/rfc6749/endpoints/metadata.py
```diff
@@ -68,7 +69,7 @@ def validate_metadata(self, array, key, is_required=False, is_list=False, is_url
                 raise ValueError("key {} is a mandatory metadata.".format(key))
 
         elif is_issuer:
-            if not array[key].startswith("https"):
+            if not is_secure_transport(array[key]):
```
auvipy, line 72: can you please verify this change with unit test?

auvipy (changes requested):

luhn:
I was going off of the existing unit tests for the metadata endpoint, which don't seem to test the validations?  Likely because the validations don't result in a user-facing error but a hard `ValueError`.

Happy to whip up a quick test if you'd still like me to.

auvipy:
yes please

auvipy (approved):

PR #806, opened 2022-02-16
Add CORS support for Refresh Token Grant.

luhn:
Exactly the same functionality as #791, added to the Refresh Token Grant.


JonathanHuot (approved):
LGTM

Issue #809, opened 2022-03-18
Show Diff for Changed Scope


jdevries3133:
**Describe the feature**

When scopes change, show the differences between the new and old scopes.

**Additional context**

I experienced the output of this code path:

https://github.com/oauthlib/oauthlib/blob/b123283ba3d41acb3e787fdf68bd5907972b4bad/oauthlib/oauth2/rfc6749/parameters.py#L457-L471

I found that the output was hard to read. I was able to better debug what was happening by adding some code to this effect into oauthlib:

```diff
diff --git a/oauthlib/oauth2/rfc6749/parameters.py b/oauthlib/oauth2/rfc6749/parameters.py
index 44738bb..26d4ca8 100644
--- a/oauthlib/oauth2/rfc6749/parameters.py
+++ b/oauthlib/oauth2/rfc6749/parameters.py
@@ -459,8 +459,15 @@ def validate_token_parameters(params):
     # parameter to inform the client of the actual scope granted.
     # https://tools.ietf.org/html/rfc6749#section-3.3
     if params.scope_changed:
-        message = 'Scope has changed from "{old}" to "{new}".'.format(
-            old=params.old_scope, new=params.scope,
+        message = 'Scope has changed from "{old}" to "{new}".\nDifference:\n{diff}'.format(
+            old=params.old_scope,
+            new=params.scope,
+            diff="\n".join(
+                difflib.ndiff(
+                    sorted(params.old_scope.split(' ')),
+                    sorted(params.scope.split(' '))
+                )
+            )
         )
         scope_changed.send(message=message, old=params.old_scopes, new=params.scopes)
         if not os.environ.get('OAUTHLIB_RELAX_TOKEN_SCOPE', None):
```

Python's `difflib` makes the addition of this feature trivial. I'm sure that this implementation isn't perfect, but I'm wondering if this looks like a nice feature. I'm interested in hearing what the community thinks, and I'd be happy to write a PR based on feedback.

- Does the feature apply to OAuth1, OAuth2 and/or OIDC? **Oauth2**
- Does the feature apply to client or server side code? **Client**

auvipy:
I would like to see a draft PR on this

PR #810, opened 2022-04-01
Tests for warning behavior on oauth scope change

jdevries3133:
When Oauth scope changes, oauthlib raises a Warning with special attributes (
`token`, `old_scope`, and `new_scope`). This behavior was documented in the
original commit message from ca4811b3087f9d34754d3debf839e247593b8a39, but
there are no tests to check for this behavior. I started working on #809 and
while reading the code, I thought the extra attributes tacked onto the warning
were strange. I thought it was even more strange that this behavior wasn't
documented anywhere or tested for.

I decided not to add any documentation because there is no existing
documentation about scope changes already, and this is a very specific
situation, however I think that having the test case, and the
`TokenScopeChangedWarning` class makes it much clearer to anyone taking a
glance at the source code what the behavior is.

Plus, I didn't want to break anything while working on #809, and I became a
little suspicious when I deleted those lines (`w.token = params; w.old_scope = ...`),
and all the unit tests still passed!

### DIFF
--- a/oauthlib/oauth2/rfc6749/parameters.py
+++ b/oauthlib/oauth2/rfc6749/parameters.py
@@ -442,6 +442,13 @@ def parse_token_response(body, scope=None):
     return params
 
 
+class TokenScopeChangedWarning(Warning):
+    token: dict
+    old_scope: List[str]
+    new_scope: List[str]
+
+
+
 def validate_token_parameters(params):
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
please check the CI failures

auvipy:
https://app.travis-ci.com/github/oauthlib/oauthlib/builds/248883785 1 build failures

PR #821, opened 2022-06-15
drop python 3.6 from matrix

auvipy:


skion:
The tagline in the top-level README also mentions Python 3.6.

auvipy:
yeah that need to be addressed as well

JonathanHuot:
Cancelled as travis is abandonned.

PR #836, opened 2022-09-16
Use proper SPDX identifier

Shortfinga:


setup.py
```diff
@@ -32,7 +32,7 @@ def fread(fn):
     maintainer_email='ib.lundgren@gmail.com',
     url='https://github.com/oauthlib/oauthlib',
     platforms='any',
-    license='BSD',
+    license='BSD-3-Clause',
```
auvipy, line 35: reference to any resource for this?

Shortfinga: Well not sure what resources you mean...

## Regarding the BDS-3-Clause:
Your project seems be licensed with a BSD-3-Clause, Github detects it as such: https://github.com/oauthlib/oauthlib/blob/master/LICENSE
To be honest I haven't compared every word, but the similarity to the reference https://spdx.org/licenses/BSD-3-Clause.html seems to be there.
So I assumed you're using the BSD-3-Clause license.

## Regarding the SPDX:
SPDX is a standard of Linux foundation: https://spdx.dev/
> SPDX is an open standard for communicating software bill of material information, including provenance, license, security, and other related information. SPDX reduces redundant work by providing common formats for organizations and communities to share important data, thereby streamlining and improving compliance, security, and dependability. The SPDX specification is recognized as the international open standard for security, license compliance, and other software supply chain artifacts as ISO/IEC 5962:2021.

AFAIK it makes it much easier for automatic compliance/license checks. So somebody knows faster or automatically how to treat this software, e.g. is this Copyleft, do I have to include a copy of the license e.g...
BSD alone could mean BSD0, BSD-2-Clause, BSD-3-Clause-No-Nuclear-License... The list is quite long: https://spdx.org/licenses/

I haven't studied it in too much detail but am glad to answer more questions.

auvipy (approved):

PR #848, opened 2023-04-14
Extend generate_signed_token with extra params

tomek-at-rais:
Extend `generate_signed_token` to allow custom algorithm and headers in JWT.

### DIFF
--- a/oauthlib/common.py
+++ b/oauthlib/common.py
@@ -197,7 +197,7 @@ def generate_token(length=30, chars=UNICODE_ASCII_CHARACTER_SET):
     return ''.join(rand.choice(chars) for x in range(length))
 
 
-def generate_signed_token(private_pem, request):
+def generate_signed_token(private_pem, request, algorithm='RS256', headers=None):
     import jwt
 
     now = datetime.datetime.utcnow()
@@ -209,7 +209,7 @@ def generate_signed_token(private_pem, request):
 
     claims.update(request.claims)
 
-    token = jwt.encode(claims, private_pem, 'RS256')
+    token = jwt.encode(claims, private_pem, algorithm=algorithm, headers=headers)
     token = to_unicode(token, "UTF-8")
 
     return token


oauthlib/common.py
```diff
@@ -197,7 +197,7 @@ def generate_token(length=30, chars=UNICODE_ASCII_CHARACTER_SET):
     return ''.join(rand.choice(chars) for x in range(length))
 
 
-def generate_signed_token(private_pem, request):
+def generate_signed_token(private_pem, request, algorithm='RS256', headers=None):
```
auvipy, line 200: we will need test coverage for this change

auvipy (changes requested):

oauthlib/common.py
tomek-at-rais: @auvipy I don't see any tests for `generate_signed_token` so I can't test "this change". 

There's `signed_token_generator` usage in `oauthlib/tests/oauth2/rfc6749/test_server.py`. It would be easiest to extend `signed_token_generator` with new parameters and then use `SignedTokenEndpointTest` to add some tests for those new parameters. But technically speaking this would be testing `signed_token_generator` not `generate_signed_token`.

I'm open to suggestion on how to approach this. Also, how do I easily mock `request` parameter of `generate_signed_token` without getting into too many oauth details? This is a method from `common` so I shouldn't have to use any code that's nested under `rfc...` folders.

foode222: foode222

foode222: > @auvipy I don't see any tests for `generate_signed_token` so I can't test "this change". 
> 
> 
> 
> There's `signed_token_generator` usage in `oauthlib/tests/oauth2/rfc6749/test_server.py`. It would be easiest to extend `signed_token_generator` with new parameters and then use `SignedTokenEndpointTest` to add some tests for those new parameters. But technically speaking this would be testing `signed_token_generator` not `generate_signed_token`.
> 
> 
> 
> I'm open to suggestion on how to approach this. Also, how do I easily mock `request` parameter of `generate_signed_token` without getting into too many oauth details? This is a method from `common` so I shouldn't have to use any code that's nested under `rfc...` folders.

tomek-at-rais: Anyone? @auvipy?

auvipy: @JonathanHuot

tomek-at-rais:
I'm very sorry no-one is interested in helping me contribute to this project.

PR #850, opened 2023-05-18
Create dependency-review.yml

auvipy:


cclauss:
Please rebase so that `lint_python` will pass.

cclauss:
I have been trying this out on several Python projects and am unable to get this Action to fail.  `lint-python` is currently running https://pypi.org/project/safety which fails on known vulnerabilities (like the earlier versions of `setuptools`).  Is there a Python dependency that we could add to the project that will prove that this Actions can actually fail?

cclauss:
```bash
git checkout master -- .github/workflows/lint_python.yml  # the dash-dash are significant
git commit -am"Update .github/workflows/lint_python.yml"
```

auvipy (commented):
I think we can merge this as Dependency Review / dependency-review (pull_request)  is passing and working

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,20 @@
...
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@v2
```
auvipy, line 20:
```suggestion
        uses: actions/dependency-review-action@v3
```

cclauss:
> we can merge this as Dependency Review / dependency-review (pull_request) is passing

The failing `lint_python` test does not worry me.  It is failing because `safety` found a vulnerability in `setuptools` which was fixed in:
* #854 (merged)

As [stated above](https://github.com/oauthlib/oauthlib/pull/850#issuecomment-1691673342), my worry is that there is no proof that `dependency-review` can ever **_fail_** on a Python project.

.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,20 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v3
```
cclauss, line 18:
```suggestion
        uses: actions/checkout@v4
```
https://github.com/actions/checkout/releases

JonathanHuot:
When the setuptools version is added in requirements.txt, it is detecting it correctly and merging is blocked.
capture

Note that different lang has different checks, according to https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems

I suppose it becomes trickier when dependencies are installed in GitHub Actions without versions, as the Dependency API won't have a way to check changes.. 

I think we're good to proceed even tho it seems a duplicate of `safety check` which is also running it.

cclauss:
#871 fixes some of the remaining issues.

PR #853, opened 2023-08-11
Ensure that `request.client_id` is set during Refresh Token Grant.

luhn:
If a confidential client authenticates via Authorization header, `request.client_id` will be `null`.  This change ensures that `request.client_id` is always set.

Authorization Code Grant has similar behavior, see 335d9e8b427163bc64c18cb7ad54e609fcd84b23

### DIFF
--- a/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py
+++ b/oauthlib/oauth2/rfc6749/grant_types/refresh_token.py
@@ -101,6 +101,9 @@ class RefreshTokenGrant(GrantTypeBase):
             if not self.request_validator.authenticate_client(request):
                 log.debug('Invalid client (%r), denying access.', request)
                 raise errors.InvalidClientError(request=request)
+            # Ensure that request.client_id is set.
+            if request.client_id is None and request.client is not None:
+                request.client_id = request.client.client_id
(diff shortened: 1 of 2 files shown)


oauthlib/oauth2/rfc6749/grant_types/refresh_token.py
```diff
@@ -101,6 +101,8 @@ def validate_token_request(self, request):
             if not self.request_validator.authenticate_client(request):
                 log.debug('Invalid client (%r), denying access.', request)
                 raise errors.InvalidClientError(request=request)
+            # Ensure that request.client_id is set.
+            request.client_id = request.client_id or request.client.client_id
```
auvipy, line 105: can you please add some test to verify the change and it won't introduce any regression?

auvipy (changes requested):

oauthlib/oauth2/rfc6749/grant_types/refresh_token.py
cclauss: Also, please rebase because `lint_python` should now pass.

luhn: Okay, added a test and rebased.

auvipy (approved):

PR #858, opened 2023-08-26
CI: Only attempt upload for upstream

hugovk:
Add an extra conditional because there's no point attempting upload from forks because they'll fail as they don't have the credentials.

# Before

image

> Notice: Attempting to perform trusted publishing exchange to retrieve a temporary short-lived API token for authentication against https://upload.pypi.org/legacy/ due to __token__ username with no supplied password field

https://github.com/hugovk/oauthlib/actions/runs/5984901609

Also remove the double curly brackets, they prevent the `if` statement from being evaluated properly.

# After

image

https://github.com/hugovk/oauthlib/actions/runs/5985055507

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   pypi-publish:
     if: |
+      github.repository_owner == 'oauthlib' &&
       ${{ github.event.workflow_run.conclusion == 'success' }} &&
       ${{ github.ref_type == 'tag' }}
     name: Upload release to PyPI


cclauss (approved):

.github/workflows/python-publish.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   pypi-publish:
     if: |
+      github.repository_owner == 'oauthlib' &&
```
auvipy, line 10: is the owner right permission? or it should be member or something?

auvipy (changes requested):

.github/workflows/python-publish.yml
hugovk: Yes, see the docs:

> The repository owner's username. For example, `octocat`.

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

And see this similar PR: https://github.com/python-attrs/attrs/pull/1179

auvipy (approved):

PR #863, opened 2023-09-02
Migrate metadata from setup.py to setup.cfg

### LINKED ISSUE #862: Adopt `build` because running `setup.py` is deprecated for security reasons
https://setuptools.pypa.io/en/latest/userguide/quickstart.html#setup-py
> To avoid executing arbitrary scripts and boilerplate code, we are transitioning into a full-fledged `setup.cfg` to declare your package information instead of running `setup()`.

See **warning** at https://setuptools.pypa.io/en/latest/deprecated/commands.html
> running `python setup.py` directly as a script is considered **deprecated**. This also means that users should avoid running commands directly via `python setup.py <command>`.

% `git grep setup\\.py`
```
.github/workflows/python-publish.yml:      run: python setup.py build
.github/workflows/python-publish.yml:      run: python setup.py sdist bdist_wheel
CHANGELOG.rst:* Several minors updates to setup.py and tox
docs/release_process.rst:they lock the oauthlib version in ``setup.py`` and release anyway.
tox.ini:# as of today, RTD is using python3.7 and doesn't run "setup.py install"
```

https://pypa-build.readthedocs.io

Also, use https://pypi.org/project/setuptools-py2cfg to eliminate or greatly reduce `setup.py`.  This could be a step towards using https://pypi.org/project/pyproject-migrator to get to replace `setup.*` with a `pyproject.toml` file.

cclauss:
Closes #862
* #862

Created using https://pypi.org/project/setuptools-py2cfg plus manual modifications.

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use
> We recommend users expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).
---
___After___ this pull request is tested and merged, I can do a follow-up PR like...

### PEP 621: Migrate settings from setup.cfg into pyproject.toml
Migrate settings from `setup.cfg` into `pyproject.toml` using [ini2toml](https://pypi.org/project/ini2toml) to do the file conversion and running [pyproject-fmt](https://pypi.org/project/pyproject-fmt) and then [validate-pyproject](https://github.com/abravalheri/validate-pyproject) in pre-commit to validate the results.


setup.cfg
```diff
@@ -1,12 +1,59 @@
 [metadata]
+name = oauthlib
+version = attr: oauthlib.__version__
+author = The OAuthlib Community
+author_email = idan@gazit.me
```
auvipy, line 5: we can use the email of two current active maintainer here

setup.cfg
```diff
@@ -1,12 +1,59 @@
...
+description = A generic, spec-compliant, thorough implementation of the OAuth request-signing logic
+license = BSD-3-Clause
 license_files = LICENSE
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+maintainer = Ib Lundgren
```
auvipy, line 11:
```suggestion
maintainer = Asif Saif Uddin, Jonathan Huot
```

auvipy (changes requested):

.github/workflows/python-publish.yml
```diff
@@ -22,12 +22,10 @@ jobs:
...
+        python-version: '3.11'
     - name: Install prereq
-      run: pip install wheel
-    - name: Build python package
-      run: python setup.py build
-    - name: Package python package
-      run: python setup.py sdist bdist_wheel
+      run: pip install build wheel
+    - name: Build Python package
+      run: python -m build
```
auvipy, line 29: do this provide both sdist bdist_wheel at once?

setup.cfg
```diff
@@ -1,12 +1,59 @@
...
+version = attr: oauthlib.__version__
+author = The OAuthlib Community
+author_email = idan@gazit.me
+description = A generic, spec-compliant, thorough implementation of the OAuth request-signing logic
+license = BSD-3-Clause
 license_files = LICENSE
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+maintainer = Asif Saif Uddin, Jonathan Huot
+maintainer_email = ib.lundgren@gmail.com
```
auvipy, line 12:
```suggestion
maintainer_email = auvipy@gmail.com
```

auvipy (changes requested):

.github/workflows/python-publish.yml
cclauss: https://pypa-build.readthedocs.io

auvipy: i saw that before, but just asking out of curiosity, do that provide same functionalities in a simpler new way?

setup.cfg
cclauss: @JonathanHuot Your call on changing the Author?

setup.cfg
cclauss: @JonathanHuot Do you want your email address published as a Maintainer?

.github/workflows/python-publish.yml
cclauss: `build` is the PyPA default tool for doing builds without a `setup.py` but there are many other ways to do builds.  I suggest we look at the build results to see if they fit our purposes.

PR #865, opened 2023-09-14
ServiceApplicationClient: Add extra_jwt_headers

lukaslihotzki-f:
This is needed to use the [Zitadel token endpoint](https://zitadel.com/docs/apis/openidoauth/authn-methods) that needs an extra `"kid"` header with ServiceApplicationClient.

I named the parameter `extra_jwt_headers` (not just `extra_headers`) to disambiguate it from HTTP headers.


auvipy (changes requested):
can you add some unit tests for this? it is needed

auvipy (commented):
lets wait for the CI

lukaslihotzki-f:
All checks have passed. @auvipy, can you merge it?

oauthlib/oauth2/rfc6749/clients/service_application.py
```diff
@@ -68,6 +68,7 @@ def prepare_request_body(self,
                              audience=None,
                              expires_at=None,
                              issued_at=None,
+                             extra_jwt_headers=None,
```
auvipy, line 71: is it needed for any specific specification or for a particular API/use case only?

lukaslihotzki-f: Standard header parameters are defined in [RFC 7515](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1). In this document, `"kid"` is defined, but also other parameters that may be meaningfully used here.

auvipy:
lets wait for review from another maitainer

PR #866, opened 2023-10-24
Make UtilsTests.test_filter_params Python 3.13+ compatible

hroncok:
Since Python 3.13.0a1, docstrings are automatically dedented. See https://github.com/python/cpython/issues/81283 and https://docs.python.org/3.13/whatsnew/3.13.html#other-language-changes

As a result, using a docstring with leading space as a test case breaks the test assumption.

The initial commit which introduced this test a decade ago (6c0c7914f3a57823834b1be492b307992f943629)
does not specify why testing the spaces is important.


auvipy (changes requested):
we are having lint issues

hroncok:
```
Run ruff --format=github .
error: unexpected argument '--format' found
```

auvipy (approved):

Issue #856, opened 2023-08-24
typing issues


jvanasco:
I've run into some issues adding typing to a project that uses oauthlib.

The first issue is regarding `oauthlib.common.Request`.  

The package is managing many object attributes through a dict, which creates issues with typing.  See  https://github.com/oauthlib/oauthlib/blob/master/oauthlib/common.py#L360-L401  

I initially opened a PR against the stubs here (https://github.com/python/typeshed/pull/10613), however typeshed noticed this section of code appears to be designed to handle parsed http(s) request data and should be strints.

This pattern is not shared across the library.

For example, Reference the docs:
https://github.com/oauthlib/oauthlib/blob/4a7db54f005686128102d7f7ac5c3d783c244669/docs/oauth2/endpoints/resource.rst#L31

```
# oauthlib_request has a few convenient attributes set such as
# oauthlib_request.client = the client associated with the token
# oauthlib_request.user = the user associated with the token
# oauthlib_request.scopes = the scopes bound to this token
```

And then again here, we see that `request.client` should be an object: 
 https://github.com/oauthlib/oauthlib/blob/4a7db54f005686128102d7f7ac5c3d783c244669/oauthlib/oauth2/rfc6749/grant_types/client_credentials.py#L107-L113

This is repeated often throughout the code.

It's been a few years since I've worked much with this library, but I believe the following are supposed to be objects:

* client
* user
* refresh_token

And several others should be Lists of strings.

IMHO, the Request class should have declared attributes for the objects, lists and strings that are not simply decoded from the http(s) query.

jvanasco:
Another issue with typing can only be solved with a breaking change:

This library often uses the same name for an object of different types.  This happens a lot with "token"s – which can either be a string OR an object/dict.  I think users of the library would benefit if tokens that should be objects/dicts were named appropriately.

auvipy:
I am open to any relevant improvement proposal

PR #869, opened 2023-11-20
initial idea for clarifying names of token -> token_dict and token_str

jvanasco:
Here is an initial idea to start addressing #856

Various functions accept a parameter with a name like "token", but the parameter might either be a String or a Dict depending on the function.  Sometimes two similar functions will require different input types - for example `save_bearer_token` operates on a Dict, but `validate_bearer_token` operates on a string.  To better clarify this, and aid in future typing, the names of these functions were updated to have a `_str` or `_dict` suffix.  On several functions, the type of input was not identified in the docstring and a code audit was needed to determine. 

This does not change the various attributes on `oauthlib.common.Request` - which do need to be better listed and typed.

This quick PR was done for initial feedback.  I did not run this through any tests yet.

### DIFF
--- a/docs/oauth2/grants/custom_grant.rst
+++ b/docs/oauth2/grants/custom_grant.rst
@@ -63,8 +63,8 @@ This example shows how to add a simple extension to the `Token endpoint`:
             # ..
             self.request_validator.your_custom_check(request)
 
-            token = token_handler.create_token(request)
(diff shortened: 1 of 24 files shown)


auvipy (changes requested):
the name change scheme seems to be OK but need input from more maintainers. also the related tests also should be renamed I think

PR #871, opened 2023-12-22
Coveralls parallel is True — Turn GitHub Actions green

cclauss:
https://coveralls-python.readthedocs.io/en/latest/usage/configuration.html#github-actions-support

https://coveralls.io/github/oauthlib/oauthlib

PR #872, opened 2024-01-06
lint_python.yml: ruff --output-format=github .

cclauss:
Ruff renamed this option when they added a code formatted similar to black.


cclauss:
Closing in favor of #871 which fixes ruff and codespell.

PR #874, opened 2024-02-06
fix: redirect URI should be the last parameter

barredterra:
https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2 says:

> The endpoint URI MAY include an "application/x-www-form-urlencoded" formatted (per [Appendix B](https://datatracker.ietf.org/doc/html/rfc6749#appendix-B)) query component ([[RFC3986] Section 3.4](https://datatracker.ietf.org/doc/html/rfc3986#section-3.4)), which MUST be retained when adding additional query parameters.

Query parameters added after the redirect URI may be interpreted as part of the redirect URI, resulting in an "Invalid Redirect URI" error from the authorization server. Adding `redirect_uri` as the last query parameter seems to be an easy way to avoid this.

Example:

```python
oauth = OAuth2Session(
	client_id="my-client-id",
	redirect_uri="https://example.org/oauth/callback/abc",
	scope=["my-scope"],
)
authorization_url, state = oauth.authorization_url(self.authorization_uri, access_type=offline)
authorization_url == "https://accounts.google.com/o/oauth2/v2/auth?response_type=code&client_id=my-client-id&redirect_uri=https://example.org/oauth/callback/abc&scope=my-scope&state=my-state&access_type=offline"
```

Now the authorization server might think that the redirect URI is `"https://example.org/oauth/callback/abc&scope=my-scope&state=my-state&access_type=offline"`, which does not match the registered URI `"https://example.org/oauth/callback/abc"`.

After this PR, the resulting authorization_url would be `"https://accounts.google.com/o/oauth2/v2/auth?response_type=code&client_id=my-client-id&scope=my-scope&state=my-state&access_type=offline&redirect_uri=https://example.org/oauth/callback/abc"`, so the `redirect_uri` cannot be interpreted in a wrong way.

### DIFF
--- a/oauthlib/oauth2/rfc6749/parameters.py
+++ b/oauthlib/oauth2/rfc6749/parameters.py
@@ -73,8 +73,6 @@ def prepare_grant_uri(uri, client_id, response_type, redirect_uri=None,
     params = [(('response_type', response_type)),
               (('client_id', client_id))]
 
-    if redirect_uri:
-        params.append(('redirect_uri', redirect_uri))
     if scope:
         params.append(('scope', list_to_scope(scope)))
     if state:
@@ -86,6 +84,9 @@ def prepare_grant_uri(uri, client_id, response_type, redirect_uri=None,
     for k in kwargs:
         if kwargs[k]:
             params.append((str(k), kwargs[k]))
+
+    if redirect_uri:
+        params.append(('redirect_uri', redirect_uri))
 
     return add_params_to_uri(uri, params)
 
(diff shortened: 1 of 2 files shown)


auvipy (changes requested):
I started the CI. is it possible to adjust the unit test to see if any regression happens?

barredterra:
Please carefully verify that the problem and fix are valid, since this is my first contribution here and I don't really know what I'm doing. 😄

PR #876, opened 2024-08-04
Update create_code_verifier to output the proper length

shawnz:
Previously, `create_code_verifier` would call `secrets.token_urlsafe(nbytes=length)` to produce the code verifier, but this leads to incorrect behaviour because the string returned by `token_urlsafe` doesn't have length `nbytes`. Instead it has length `nbytes * (4/3)` since the output is base64-encoded.

This means that, for example, you could call `create_code_verifier(length=128)` and you would get a result with length 171, which exceeds the 128-character maximum length of the code verifier allowed by the spec. But this wouldn't be caught by the checks in the code, since they only check the input length parameter and not the actual output length.

This PR fixes the issue by using the `generate_token` utility function instead, which does output strings of the exact length you specify. It also adds a test to make sure the length of the outputted string is equal to the specified length.

### DIFF
--- a/oauthlib/oauth2/rfc6749/clients/base.py
+++ b/oauthlib/oauth2/rfc6749/clients/base.py
@@ -490,7 +490,7 @@ class Client:
         if not length <= 128:
             raise ValueError("Length must be less than or equal to 128")
 
-        allowed_characters = re.compile('^[A-Zaa-z0-9-._~]')
+        allowed_characters = re.compile('^[A-Za-z0-9-._~]')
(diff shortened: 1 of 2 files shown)


thedrow (approved):

auvipy (changes requested):
we have to make the CI build green

shawnz:
Hi @auvipy, see #878 where I make an attempt to fix the CI errors. I did it in a separate PR since they are not related to this change.

auvipy:
can you please rebase? merged the PR

shawnz:
@auvipy done, ty for looking into this

auvipy (approved):

PR #878, opened 2024-08-12
Fix CI Errors

shawnz:
This fixes the CI errors that are currently causing my other PR #876 to fail.

There were a few issues:

- Ruff now requires an explicit `check` command in the command line
- PyJWT now raises an InvalidKeyError in a situation where it previously raised a ValueError
- Several Ruff linting rules were being violated, most of which I fixed with ruff check --fix, and the remaining I marked as ignored

I also cleaned up some warning detection code in one of the tests and silenced one of the expected DeprecationWarnings which was previously leaking through to the output

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -13,7 +13,7 @@ jobs:
       - run: pip install black codespell mypy pytest ruff safety
       - run: ruff check --output-format=github .
       - run: black --check . || true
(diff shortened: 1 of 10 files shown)


shawnz:
Note there are also many black and mypy errors, but they are not required to pass CI. Should I fix them anyway?

auvipy (approved):

PR #879, opened 2024-08-27
Guard ui_locales.split()

jaap3:
`request.ui_locales` might already be a list, only split the value if the `split` method is available.

I'm creating this PR because I ran into an issue with `django-oauth-toolkit`, I've added a workaround there:

https://github.com/jazzband/django-oauth-toolkit/pull/1469

Since I'm unsure if this is a bug in `oauthlib`, or is caused by unexpected use by `django-oauth-toolkit` I'm suggesting this change here as well.

### DIFF
--- a/oauthlib/openid/connect/core/grant_types/base.py
+++ b/oauthlib/openid/connect/core/grant_types/base.py
@@ -310,11 +310,15 @@ class GrantTypeBase:
             msg = "Session user does not match client supplied user."
             raise LoginRequired(request=request, description=msg)
 
+        ui_locales = request.ui_locales if request.ui_locales else []
+        if hasattr(ui_locales, 'split'):
+            ui_locales = ui_locales.strip().split()
+
         request_info = {
             'display': request.display,
             'nonce': request.nonce,
             'prompt': prompt,
-            'ui_locales': request.ui_locales.split() if request.ui_locales else [],
+            'ui_locales': ui_locales,
             'id_token_hint': request.id_token_hint,
             'login_hint': request.login_hint,
             'claims': request.claims


oauthlib/openid/connect/core/grant_types/base.py
```diff
@@ -310,11 +310,15 @@ def openid_authorization_validator(self, request):
             msg = "Session user does not match client supplied user."
             raise LoginRequired(request=request, description=msg)
 
+        ui_locales = request.ui_locales if request.ui_locales else []
+        if hasattr(ui_locales, 'split'):
```
9128305, line 314: isinstance check should be faster

jaap3: Maybe, but I'm just trying to mirror the code style already used in this file (see line 290)

9128305: I see...

auvipy (changes requested):
we will need tests to verify the proposed change as well

auvipy (changes requested):
please add/modify unit tests to verify this

tests/openid/connect/core/endpoints/test_openid_connect_params_handling.py
```diff
@@ -28,7 +28,8 @@ def setUp(self):
             'redirect_uri': 'https://a.b/cb',
             'response_type': 'code',
             'client_id': 'abcdef',
-            'scope': 'hello openid'
+            'scope': 'hello openid',
+            'ui_locales': 'en-US'
```
jaap3, line 32: By adding this line the `test_authorization_endpoint_handles_prompt` will fail without the patch.

I also added a line to `def test_oidc_params_preservation(self):` to verify `ui_locales` is parsed.

jaap3 (commented):
@auvipy I added/updated the tests.

auvipy (approved):

PR #885, opened 2024-10-14
Allow user code to be configured

duzumaki:
Oauth2 doesn't specify the format of the user_code token but ideally it should be something human and device display friendly. e.g something short like ME3-2FD

So let's allow the this code to be configured based on the needs of the device instead.

The current token generation logic is now the default behaviour

e.g response

```
{
    "verification_uri": "xxx",
    "expires_in": 1800,
    "user_code": "M3D-DGF" (instead of a non human friendly string like with device_code),
    "device_code": "9Y4cH5zUXutDJcAwXJZoRjJnxNn36C"
}

```

@ maintainers, I'm currently working on a pr for [this](https://github.com/jazzband/django-oauth-toolkit/issues/962) issue in DOT so I have tested this with a view i've implemented on a branch there, that a custom authentication server I'm working on is using.


oauthlib/oauth2/rfc8628/endpoints/device_authorization.py
```diff
@@ -49,11 +51,11 @@ def __init__(
         :param verification_uri_complete: a string of a function that can be called with `user_data` as parameter
         """
         self.request_validator = request_validator
+        self.user_code_generator = user_code_generator
         self._expires_in = expires_in
         self._interval = interval
         self._verification_uri = verification_uri
         self._verification_uri_complete = verification_uri_complete
-        self._interval = interval
```
duzumaki, line 56: This is assigned twice

oauthlib/oauth2/rfc8628/endpoints/device_authorization.py
```diff
@@ -39,6 +40,7 @@ def __init__(
         expires_in=1800,
         interval=None,
         verification_uri_complete=None,
+        user_code_generator: Callable[[None], str] = None
```
duzumaki, line 43: I thought it would be nice to introduce typing to this repo to make objects easier to understand without having to check during run time

oauthlib/oauth2/rfc8628/endpoints/device_authorization.py
```diff
@@ -32,14 +34,7 @@ class DeviceAuthorizationEndpoint(BaseEndpoint):
...
 
-    def __init__(
-        self,
-        request_validator,
-        verification_uri,
-        expires_in=1800,
-        interval=None,
-        verification_uri_complete=None,
-    ):
+    def __init__(self, request_validator, verification_uri, expires_in=1800, interval=None, verification_uri_complete=None, user_code_generator: Callable[[None], str] = None):
```
auvipy, line 37: I think the previous hanging style format looks better and seems more readable.

auvipy (changes requested):
can you also update the docs with example please?

oauthlib/oauth2/rfc8628/endpoints/device_authorization.py
duzumaki: I agree but this is what ruff formatted it to.

I'll play around with the ruff config

duzumaki: @auvipy I locally changed the max line length to 99 for ruff. The 255 is too high. Would need a seperate pr to format the entire repo

[commit here](https://github.com/oauthlib/oauthlib/pull/885/commits/6862c41f3994f770c3a1fbdb399a689f6e98a520)

duzumaki:
@auvipy Added device docs

oauthlib/oauth2/rfc8628/endpoints/device_authorization.py
duzumaki: @auvipy 
Ideally the ruff config should have a lower max line length but running it on the code base leads to too many violations in this repo so just changed the files related to this pr in regards to formatting

[commit here](https://github.com/oauthlib/oauthlib/pull/885/commits/51ddcf652642c29d4b70fc07c7258d5f08509101)

auvipy (approved):

auvipy:
merging this, but feel free to contribute back if any edge case arises

PR #891, opened 2025-01-24
Device flow: Pass verification_uri_complete to endpoint + pass Server kwargs to DeviceCodeGrant to allow validators to be setup with more flexibility

duzumaki:
Note to reviewer: pr is best reviewed commit by commit

While implementing [this](https://github.com/jazzband/django-oauth-toolkit)
I realised two things are missing

1. `verification_uri_complete` isn't being passed to the endpoint
2. The server kwargs aren't being passed to the device code grant

### Explanation of 2:
oauthlib assumes code like so for adding custom validators (or rather, examples in the code base do this but the constructor of a grant allows for generic kwargs to be passed in to set it up)

```
auth = DeviceCodeGrant(validator)

auth.custom_validators.pre_token.append(pre_token_validator)
auth.custom_validators.post_token.append(post_token_validator)
```

However there are cases where one would want more flexibility. What do I mean flexibility? All grant's inherit from `GrantTypeBase` and have a [constructor](https://github.com/duzumaki/oauthlib/blob/a4af43e35ba98ad84e7f6df09d541a6d1e0c2135/oauthlib/oauth2/rfc6749/grant_types/base.py#L82) like:

```
    def __init__(self, request_validator=None, **kwargs):
        self.request_validator = request_validator or RequestValidator()
        self._setup_custom_validators(kwargs)

    def _setup_custom_validators(self, kwargs):
        post_auth = kwargs.get('post_auth', [])
        post_token = kwargs.get('post_token', [])
        pre_auth = kwargs.get('pre_auth', [])
        pre_token = kwargs.get('pre_token', [])
        self.custom_validators = ValidatorsContainer(post_auth, post_token,
                                                     pre_auth, pre_token)
```

One might want to pass in any combination of the custom validators to the `Server` that uses the Grant object as a generic dict of [server kwargs ](https://github.com/duzumaki/django-oauth-toolkit/blob/af5d436e16802c48b4c37910c450640f02a7b6e1/build/lib/oauth2_provider/views/mixins.py#L77)(like is done in [django-oauth-toolkit](https://github.com/jazzband/django-oauth-toolkit) in order to abstract oauthlib away)

By allowing this it makes it easier to set up the validator checks

#### But what were you even trying to do in the first place?
For the device code this means being able to pass in a validation callable in a more flexible manner to assign `oauthlib.common.Request's` user attribute before the access token is made and saved allowing to associate the device user with the access token user 

Screenshot showing access tokens having users set from the device flow and `verification_uri_complete`(https://datatracker.ietf.org/doc/html/rfc8628#section-3.2) coming back from the authorization stage
image
image

### DIFF
--- a/docs/oauth2/endpoints/device.rst
+++ b/docs/oauth2/endpoints/device.rst
@@ -22,12 +22,21 @@ the device authorization endpoint.
     from your_validator import your_validator
     verification_uri = "https://example.com/device"
 
+    # verification_uri_complete can either be a callable that receives user code as an arg
+    # or a string (e.g verification_uri_complete = "https://example.com/device=1234")
+    verification_uri_complete = lambda user_code: f"https://example.com/device={user_code}"
+
     def user_code():
        # some logic to generate a random string...
        return "123-456"
 
     # user code is optional
-    server = DeviceApplicationServer(your_validator, verification_uri, user_code)
+    server = DeviceApplicationServer(
+        request_validator=your_validator,
+        verification_uri=verification_uri,
+        verification_uri_complete=verification_uri_complete,
+        user_code=user_code
+    )
 
     headers, data, status = server.create_device_authorization_response(request)
 
--- a/oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
+++ b/oauthlib/oauth2/rfc6749/endpoints/pre_configured.py
@@ -57,7 +57,7 @@ class Server(
         self.password_grant = ResourceOwnerPasswordCredentialsGrant(request_validator)
         self.credentials_grant = ClientCredentialsGrant(request_validator)
         self.refresh_grant = RefreshTokenGrant(request_validator)
-        self.device_code_grant = DeviceCodeGrant(request_validator)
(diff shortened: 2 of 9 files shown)


duzumaki:
hmm why is ruff reported on lines unrelated to my change?

duzumaki:
hmm why is ruff reporting on lines unrelated to my change?
it also isn't failing locally?

duzumaki:
@auvipy doesn't let me add reviewers so tagging

auvipy (changes requested):
shouldn't we also update relevant docs?

duzumaki:
@auvipy it's already documented :)

https://github.com/oauthlib/oauthlib/blob/028c563eb6af65426df45ba4bbfc771974e9594e/docs/oauth2/endpoints/device.rst#L47

duzumaki:
@auvipy  udpated [here](https://github.com/oauthlib/oauthlib/pull/891/commits/5aa758c08a0d72c483778ab7d0dd26df983b8411)

auvipy (approved):

PR #902, opened 2025-05-21
docs: add django-allauth to available options

pennersr:
For reference, see: https://docs.allauth.org/en/dev/idp/openid-connect/index.html


auvipy (approved):
thanks!

Issue #906, opened 2025-06-18
Issue with version 3.3.0 and other libraries


nodiz:
**Describe the bug**

`oauthlib==3.3.0` raises  `ValueError: expires_int must be an int` inside `oauthlib.oauth2.rfc6749.parameters.parse_expires()` whenever the access-token dict contains an `expires_in` value that is **not already an `int`**.  
Many real-world providers (Google, Azure AD, GitLab, Reddit, …) return `"expires_in": "3599"` (a JSON string), and earlier oauthlib versions silently coerced this to `int`.  
The change shipped in 3.3.0 breaks every downstream library that passes the raw token through, e.g. `requests-oauthlib`, `Flask-Dance`, Azure’s `msrest`, etc.  example: https://github.com/Azure/azure-sdk-for-python/issues/41635

---

**How to reproduce**

```python
# works on 3.2.2, crashes on 3.3.0
from oauthlib.oauth2 import WebApplicationClient

WebApplicationClient(
    client_id="dummy",
    token={"access_token": "xyz", "expires_in": "3600"}  # ← str on purpose
)
```

Additional context
	•	OAuth version: OAuth 2 (authorization-code & installed-app flows)
	•	Code type: Client
	•	Provider(s): Google, Azure AD, GitLab, Reddit (all return expires_in as JSON string)

A simple pin to oauthlib==3.2.2 restores functionality, so this appears to be a regression introduced in 3.3.0’s new type-check logic in parse_expires().

nodiz:
Hi Jonathan, thanks for digging in—and sorry my first snippet was inconclusive.

Could you please try the constructor with the following **expires_in** value?

```python
from oauthlib.oauth2 import WebApplicationClient

WebApplicationClient(
    "dummy",
    {"access_token": "xyz",
     "token_type": "Bearer",
     "expires_in": "3600.0"} 
)

WebApplicationClient(
    "dummy",
    {"access_token": "xyz",
     "token_type": "Bearer",
     "expires_in": "3600s"}
)
```

Both calls pass silently on 3.2.2 but raise
ValueError: expires_int must be an int on 3.3.0.


Real-world example from my Flask app (after Google login):
{
  "expires_at": 1750323582.170817,
  "expires_in": 3588.03265
}

Unless there’s a strong reason to forbid numeric strings/floats, keeping the checks relaxed would preserve compatibility with existing client stacks.

nodiz:
@JonathanHuot 
[...]

PR #907, opened 2025-06-18
Fix 3.3.0 regression of expires_in

JonathanHuot:
Support https://github.com/oauthlib/oauthlib/issues/906 discussion.


oauthlib/oauth2/rfc6749/parameters.py
```diff
@@ -500,9 +500,9 @@ def parse_expires(params):
                 # Attempt to convert to int
                 expires_in = int(params.get('expires_in'))
             except ValueError:
-                raise ValueError("expires_int must be an int")
+                raise ValueError("expires_in must be an int")
```
auvipy, line 503: interesting typo it was!

JonathanHuot:
Release has been deployed with success.
However, the publish job still have two issues:
- Duplicate push/pull_request triggers twice the publish jobs
- Errr no tags has been set, but the publish job has been triggered! this is not expected

PR #914, opened 2025-06-30
PEP 621: Migrate from setup.{py, cfg} to pyproject.toml

cclauss:
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
* https://peps.python.org/pep-0621

Migrate `setup.py` to `setup.cfg` using [setuptools-py2cfg](https://pypi.org/project/setuptools-py2cfg) plus manual modifications.  Then migrate `setup.cfg` to `pyproject.toml` using [ini2toml](https://pypi.org/project/ini2toml) to do the file conversion and running [pyproject-fmt](https://pypi.org/project/pyproject-fmt) and then [validate-pyproject](https://github.com/abravalheri/validate-pyproject) to validate the results.

```bash
uvx --with=oauthlib setuptools-py2cfg > setup.cfg
uvx 'ini2toml[full]' setup.cfg -o pyproject.toml
git add pyproject.toml
uvx pyproject-fmt pyproject.toml
uvx 'validate-pyproject[all]' pyproject.toml
migrate ruff.toml to pyproject.toml
ruff check
migrate requirements*.txt to pyproject.toml  # Tox still needs them :-(
git rm setup.py
git rm setup.cfg
git rm ruff.toml
edit GitHub Actions
```

Codespell and mypy settings (in GitHub Action) and  tox.ini could also be migrated.


cclauss:
@webknjaz, your review would be greatly appreciated, esp. on the PyPI release bits.

.github/workflows/lint_python.yml
```diff
@@ -15,8 +15,7 @@ jobs:
       - run: pip install codespell mypy pytest ruff safety
       - run: ruff check --output-format=github .
       - run: codespell --ignore-words-list="implementor,mimiced,provicers,re-use,THIRDPARTY,assertIn"  # --skip="*.css,*.js,*.lock"
-      - run: pip install -r requirements-test.txt
-      - run: pip install --editable .
+      - run: pip install --editable ".[test]"
```
webknjaz, line 18: I'm in the camp of not abusing extras for development dependencies. They are public API and exposing `pip install 'oauthlib [test]'` to the end-users is bad.

Either requirement files or dependency groups describe virtualenv state, not core packaging metadata.

.github/workflows/python-build.yml
```diff
@@ -1,6 +1,14 @@
...
 run-name: Run Tests by ${{ github.actor }}
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+  release:
+    types: [published]
+  workflow_dispatch:
+permissions:
+  id-token: write
```
webknjaz, line 10: Giving OIDC to random software in all the jobs is dangerous. I filed an issue about the problem in the publishing job. This makes the exposure even worse.

.github/workflows/python-build.yml
```diff
@@ -88,4 +96,5 @@ jobs:
     - name: Check python package
       run: twine check dist/*
     - name: Publish package distributions to PyPI
+      if: github.event_name == 'release' && github.event.action == 'published'
```
webknjaz, line 99: In the issue I filed yesterday, I outline that the entire job has to be remade, split into parts with privileges scoped + the job conditional is broken. Once that's done, this clause would only be necessary in the job level.

pyproject.toml
```diff
@@ -0,0 +1,164 @@
...
+]
+dependencies = [
+  "blinker==1.4",
+  "cryptography>=3",
+  "pyjwt>=2,<3",
+]
+optional-dependencies.rsa = [ "cryptography>=3" ]
+optional-dependencies.signals = [ "blinker>=1.4" ]
+optional-dependencies.signedtoken = [ "cryptography>=3", "pyjwt>=2,<3" ]
+optional-dependencies.test = [ "pytest>=4", "pytest-cov>=2.6", "pytest-subtests" ]
```
webknjaz, line 45: As I mentioned earlier, this would be fine as dependency groups or requirement files. But not public extras recorded in the core packaging metadata.

ruff.toml
webknjaz, line 1: I find dumping multiple configs into `pyproject.toml` problematic because it's difficult to maintain when syncing updates across many projects, following the principles of jaraco/skeleton or similar ideas.

.github/workflows/python-build.yml
JonathanHuot: This is not necessary to have ` release published` trigger as well, since currently the tag is the main trigger for uploading a package to pypi. Creating the github release has to be done manually though.

.github/workflows/python-build.yml
cclauss: Putting a * in front of a PR link makes it more readable..
* #915

cclauss: I have reverted all but one change to this file to leave room for #915 to do it correctly.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: Implementation",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Software Development :: Libraries :: Python Modules",
+]
+dependencies = [
+  "blinker==1.4",
+  "cryptography>=3",
```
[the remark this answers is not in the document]
auvipy: what about this?

.github/workflows/lint_python.yml
webknjaz: Yes, it does. `[test]` references an extra that is included as a part of the core packaging metadata. Now that it doesn't leak into the public space, you need to reference it through `--group`: https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-group

This will work as long as pip is modern: https://pip.pypa.io/en/stable/news/#v25-1.

setup.cfg
```diff
@@ -1,12 +0,0 @@
...
-
-[isort]
-combine_as_imports = true
-default_section = THIRDPARTY
-include_trailing_comma = true
-known_first_party = oauthlib
-known_tests = tests
-sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,TESTS,LOCALFOLDER
-line_length = 79
-multi_line_output = 5
```
webknjaz, line 12: Does this need to be preserved in an `.isort.cfg`?

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Software Development :: Libraries :: Python Modules",
+]
+optional-dependencies.rsa = [ "cryptography>=3" ]
+optional-dependencies.signals = [ "blinker>=1.4" ]
+optional-dependencies.signedtoken = [ "cryptography>=3", "pyjwt>=2,<3" ]
+urls = { Homepage = "https://github.com/oauthlib/oauthlib" }
+
+[dependency-groups]
+test = [ "pytest>=4", "pytest-cov>=2.6", "pytest-subtests" ]
```
webknjaz, line 43: It's usually nicer to keep such things multi-line. Version bump diffs will look better.
```suggestion
test = [
  "pytest >= 4",
  "pytest-cov >= 2.6",
  "pytest-subtests",
]
```

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: Implementation",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Topic :: Software Development :: Libraries :: Python Modules",
+]
+optional-dependencies.rsa = [ "cryptography>=3" ]
+optional-dependencies.signals = [ "blinker>=1.4" ]
+optional-dependencies.signedtoken = [ "cryptography>=3", "pyjwt>=2,<3" ]
```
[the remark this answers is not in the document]
cclauss: Let's revisit that in a separate PR.  I find it to be unintuitive.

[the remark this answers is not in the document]
JonathanHuot: Agree - recursive deps seems more prone to error and harder to read & understand. Let's keep `cryptography` in both groups.

pyproject.toml
```diff
@@ -0,0 +1,57 @@
...
+  "blinker==1.4",
+  "cryptography>=3",
+  "pyjwt>=2,<3",
+]
+optional-dependencies.rsa = [ "cryptography>=3" ]
+optional-dependencies.signals = [ "blinker>=1.4" ]
+optional-dependencies.signedtoken = [ "cryptography>=3", "pyjwt>=2,<3" ]
+urls = { Homepage = "https://github.com/oauthlib/oauthlib" }
+
+[dependency-groups]
```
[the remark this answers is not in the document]
webknjaz: @JonathanHuot your assignment is accurate. Copilot is drunk.

pyproject.toml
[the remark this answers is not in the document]
webknjaz: @cclauss I don't think having a self-reference has any effect on the dependency groups. It only deduplicates the contents of extras here.

PR #929, opened 2025-11-27
Improve Content-Type Validation for Token Endpoint Requests

Kris-Val:
This change introduces explicit validation of the Content-Type header for token endpoint requests. When the request’s Content-Type does not match the expected value, it now raises a clear and descriptive error.

[Issue #928 – oauthlib/oauthlib](https://github.com/oauthlib/oauthlib/issues/928)


auvipy (changes requested):

auvipy (commented):
not sure why the CI is not running

auvipy (changes requested):
can you please fix the lint errors?

PR #947, opened 2026-04-22
fix(errors): prefer leaf exception classes in raise_from_error for shared error codes

### LINKED ISSUE #776: oauth2 `raise_from_error` raises random error for `invalid_request`
When an OAuth2 call fails, `raise_from_error` loops through all defined exceptions in `errors.py` to find one that matches the returned error code. However, the `invalid_request` error code is defined on a base class `InvalidRequestFatalError` that is implemented by several specific errors. The result is that, when an `invalid_request` happens, a random class is raised to the calling client.

For example, I hade the following happen:
    
    oauthlib.oauth2.rfc6749.errors.InvalidClientIdError: (invalid_request) Missing required parameter: refresh_token

(which is obviously not an `InvalidClientIdError`)

armorbreak001:
## Summary

When multiple OAuth2 error classes share the same error code (e.g. `invalid_request`), `raise_from_error` would raise whichever class it encountered first during `inspect.getmembers` iteration. This could produce misleading exceptions — for instance, raising `InvalidClientIdError` for an `invalid_request` error that is actually about a missing parameter.

Fixes #776

## Details

The error class hierarchy has two base classes that define `error = 'invalid_request'`:

- `InvalidRequestFatalError` (fatal) with subclasses: InvalidRedirectURIError, MissingRedirectURIError, MismatchingRedirectURIError, InvalidClientIdError, MissingClientIdError
- `InvalidRequestError` (non-fatal) with subclasses: MissingResponseTypeError, MissingCodeChallengeError, MissingCodeVerifierError, UnsupportedCodeChallengeMethodError

Since all these classes share the same error code, the original code would raise a semi-random one depending on iteration order. The fix collects all matching classes and filters to prefer leaf (most specific) classes over base classes, so the caller gets a more accurate exception type.

### DIFF
--- a/oauthlib/oauth2/rfc6749/errors.py
+++ b/oauthlib/oauth2/rfc6749/errors.py
@@ -398,8 +398,9 @@ def raise_from_error(error, params=None):
         if cls.error == error
     ]
     # When multiple classes share the same error code (e.g. invalid_request),
-    # prefer leaf classes over base classes to avoid raising a misleading
-    # exception like InvalidClientIdError for an unrelated invalid_request.
+    # filter out base classes if more specific subclasses also match. If
+    # multiple leaf classes still match, the later selection remains dependent
+    # on inspect.getmembers ordering.
     if len(matching) > 1:
         matching_names = {cls.__name__ for cls in matching}
         matching = [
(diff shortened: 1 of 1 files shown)


auvipy (changes requested):
is this pr related https://github.com/oauthlib/oauthlib/pull/945 ?

PR #964, opened 2026-07-30
fix: use 128-bit entropy in generate_nonce

### LINKED ISSUE #946: Nonce would be more likely to be unique if more random bits replaced the timestamp
After running into a duplicate nonce error in the wild, when looking into how the nonce is generated I noticed that it is concatenating 64 random bits and the current integer timestamp. It is stated in the code that this is intended to reduce the likelihood for collisions. However, according to [RFC 5849 section 3.3](https://datatracker.ietf.org/doc/html/rfc5849#section-3.3):

> The nonce value MUST be unique across all requests with the same timestamp, client credentials, and token combinations.

In other words, according to the spec the nonce only needs to be unique *for a given integer timestamp*, which means I wouldn't expect adding the timestamp to reduce the chances of collision.

I don't know if there are considerations apart from this particular spec, but one thought is that replacing the timestamp with more random bits (e.g. adding another 32 bits if you expect timestamps to be close to 32 bits at this point) would achieve a better result while keeping the nonce the same overall length.

ltsyk:
## Summary
Increase nonce entropy from 64-bit to 128-bit random bits (issue #946).

Fixes #946


auvipy (changes requested):
please also add appropriate tests to verify this

ltsyk:
Thanks @auvipy — added a unit test that mocks `randbits` / `generate_timestamp` and asserts `randbits(128)` is used and the nonce is `str(random)+timestamp`, plus tightened the existing uniqueness check. Pushed in f7c682c.

ltsyk:
Added tests asserting `randbits(128)` and nonce composition (digits-only random prefix + timestamp suffix), including a mocked 128-bit entropy value larger than any 64-bit max.

PR #965, opened 2026-08-02
Fix device grant scope validation (#949)

### LINKED ISSUE #949: Device Grant Scopes not Validated
**Describe the bug**

`DeviceAuthorizationEndpoint.validate_device_authorization_request` does not call `validate_scopes` or `get_default_scopes()` on the request validator, unlike every other authorization endpoint in oauthlib. This means:

1. When a device omits `scope`, the resolved default scopes are never stored in the response — the `scope` key is absent from the dict returned by `create_device_authorization_response`. 
2. When a device provides `scope`, the requested scopes are never validated against the available scopes for the client.

The authorization code endpoint resolves this correctly via `base.GrantTypeBase.validate_scopes()` (rfc6749/grant_types/base.py, line 164), which falls back to `get_default_scopes()` when `request.scopes` is falsy and then calls `validate_scopes()` on the result. The device authorization endpoint has no equivalent call.

**How to reproduce**
```python
from oauthlib.oauth2 import DeviceApplicationServer
from oauthlib.oauth2.rfc6749.request_validator import RequestValidator

class MyValidator(RequestValidator):
    def validate_client_id(self, client_id, request, *args, **kwargs):
        return True

    def authenticate_client(self, request, *args, **kwargs):
        return True

    def get_default_scopes(self, client_id, request, *args, **kwargs):
        # This should be called when no scope is provided
        return ["read", "write"]

    def validate_scopes(self, client_id, scopes, client, request, *args, **kwargs):
        # This should be called to validate provided or default scopes
        return True

server = DeviceApplicationServer(
    MyValidator(),
    verification_uri="https://example.com/device",
)

headers, body, status = server.create_device_authorization_response(
    "https://example.com/device_authorization",
    http_method="POST",
    body="client_id=myclient",  # no scope provided
    headers={"Content-Type": "application/x-www-form-urlencoded"},
)

import json
response = json.loads(body)
print(response.get("scope"))  # None — get_default_scopes() was never called
```

**Expected behavior**
`validate_device_authorization_request()` should mirror the behavior of `base.GrantTypeBase.validate_scopes():`

* If scope is provided in the request, call `request_validator.validate_scopes()` to confirm the requested scopes are permitted for the client. Return an invalid_scope error if not.
* If scope is absent, call `request_validator.get_default_scopes()` to resolve the defaults, then validate them.
* Include the resolved scope in the dict returned by `create_device_authorization_response()`so that callers can store it alongside the device_code and user_code without needing to call the validator a second time themselves.

RFC 8628 §3.1 says the `scope` parameter carries the same semantics as RFC 6749 §3.3, so the same validation contract should apply.

**Additional context**

Please provide any further context here.

- Are you using OAuth1, OAuth2 or OIDC?

  OAuth2, server-side (authorization server).
 
- Are you writing client or server side code?

   Discovered while working on django-oauth-toolkit. Without this oauthlib behaviour, DOT has to work around the gap by calling get_default_scopes() manually after create_device_authorization_response() returns, requiring an extra database query to look up the Application in order to construct a sufficiently populated oauthlib.common.Request to pass to the validator. See django-oauth-toolkit PR #1692 for the workaround and its security motivation (consent integrity violation / potential privilege escalation when DEFAULT_SCOPES changes between user consent and device polling).

heyncth:
## What

`DeviceAuthorizationEndpoint.validate_device_authorization_request` never resolved or validated the requested scope, unlike every other oauthlib authorization endpoint (which go through `GrantTypeBase.validate_scopes`). Two consequences:

1. When the request omits `scope`, the default scopes were never resolved via `get_default_scopes`, and the response omitted the granted scope entirely.
2. When the request provides `scope`, the scopes were never validated against the client (`validate_scopes` was never called), so a device authorization could be issued for scopes the client is not allowed to request.

## Fix

`validate_device_authorization_request` now mirrors the `GrantTypeBase.validate_scopes` contract:

- Resolves default scopes via `get_default_scopes` when `scope` is omitted.
- Validates the resolved/requested scopes via `validate_scopes`, raising `InvalidScopeError` on failure.

`create_device_authorization_response` now includes the resolved `scope` in the response data, per RFC 8628 Section 3.2.

## Tests

- New regression tests in `tests/oauth2/rfc8628/endpoints/test_scope_validation_repro.py` covering: default scope resolution when omitted, validation failure raising `InvalidScopeError`, and requested scope echoing.
- Updated the mock validator in `tests/oauth2/rfc8628/test_server.py` for the now-enforced scope contract.

Full suite: 706 passed, 2 skipped. Ruff (pinned 0.15.2, as in pre-commit) is clean.

Closes #949


auvipy (changes requested):
closing as duplicate

PR #966, opened 2026-08-07
Add .gitattributes for cross-platform line ending normalization

chernojagne:
Add a `.gitattributes` file with standard cross-platform line ending configuration.

**What this does:**

- Sets `* text=auto` to enable Git's automatic line ending normalization for text files
- Marks binary files (images, fonts, archives, compiled objects, executables, Python bytecode, lockfiles) so they are never mangled by line ending conversion
- Marks patch and diff files as `-text` to preserve their byte sequences exactly

**Why:**

Contributors on Windows, macOS, and Linux can see different line endings in the working tree without this file. `text=auto` is Git's recommended approach for cross-platform projects — it normalizes to LF in the repo and checks out with platform-native endings. The binary markers prevent lockfiles like `uv.lock` and `poetry.lock` from being misidentified as text.

This change is purely additive and does not affect CI, tests, or existing code.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.


chernojagne:
Closing this — we're revising how we approach these contributions and would rather withdraw it than leave it open unattended. Apologies for the noise, and thanks.

